PR #161, opened 2026-04-16
feat(pagination): add ToPageAsync and WithNextToken APIs

### LINKED ISSUE #157: feat: Add ToPageAsync and WithNextToken pagination API
## Background

The provider currently has no user-facing pagination API. `Limit(n)` controls the DynamoDB evaluation budget for a single request, and `ToListAsync()` auto-follows `NextToken` internally to collect all results. There is no way for a caller to retrieve one page at a time and resume from a continuation token — the only workaround today is the shadow-property `GetExecuteStatementResponse()` escape hatch, which is not a supported API surface.

ADR-002 explicitly deferred this as "Option C". This issue implements it.

### `DynamoPage<T>` — public return type

```csharp
[Experimental]
public readonly struct DynamoPage<T>
{
    public IReadOnlyList<T> Items { get; }
    public string? NextToken { get; }  // null = no more pages
}
```

### `ToPageAsync` — explicit paging terminal

```csharp
// First page
DynamoPage<Order> page = await context.Orders
    .Where(o => o.Status == "Active")
    .ToPageAsync(limit: 25, nextToken: null);

// Subsequent pages
DynamoPage<Order> next = await context.Orders
    .Where(o => o.Status == "Active")
    .ToPageAsync(limit: 25, nextToken: page.NextToken);
```

- Terminal operator — not composable after the call
- `limit` = DynamoDB evaluation budget (items *scanned*, not *returned*)
- `nextToken` = `null` for first page, opaque string from a prior `DynamoPage<T>.NextToken`
- Translation error if combined with `Limit(n)` on the same query
- Translation error if used inside a subquery

### `WithNextToken` — cursor seeding for non-paging queries

```csharp
// Resume auto-collect from a cursor (collects all remaining results)
var remaining = await context.Orders
    .Where(o => o.Status == "Active")
    .WithNextToken(savedToken)
    .ToListAsync();

// Resume from a cursor with a one-shot budget
var chunk = await context.Orders
    .Where(o => o.Status == "Active")
    .WithNextToken(savedToken)
    .Limit(25)
    .ToListAsync();

// Compose with ToPageAsync to seed its first request
DynamoPage<Order> page = await context.Orders
    .Where(o => o.Status == "Active")
    .WithNextToken(savedToken)
    .ToPageAsync(limit: 25, nextToken: null);
```

- Non-terminal, composable
- Seeds the *first* `ExecuteStatement` request's `NextToken`
- Translation error if called more than once on the same query

## ⚠️ Critical DynamoDB Semantic

DynamoDB `Limit` is an **evaluation budget** (items scanned before filter expressions are applied), **not a result-count guarantee**.

When a query includes non-key filter predicates:

1. DynamoDB scans up to `limit` items
2. Filter expressions are applied to those scanned items
3. Only matching items are returned — which may be **fewer than `limit`**
4. `NextToken` may still be **non-null** — meaning more unscanned items exist

**The only reliable end-of-results signal is `NextToken == null`.** Do not use `Items.Count < limit` as a last-page check.

Example: a query with `limit: 10` and a non-key filter might return 3 items with a non-null `NextToken` — there are 7 more items in the budget window that did not match the filter, and more pages beyond.

This is fundamentally different from Cosmos DB's `pageSize`, which guarantees up to N *returned* items per page.

### Provider changes

- [ ] `DynamoPage<T>` readonly struct (marked `[Experimental]`)
- [ ] `ToPageAsync(int limit, string? nextToken, CancellationToken)` extension in `DynamoDbQueryableExtensions`
- [ ] `WithNextToken(string nextToken)` extension in `DynamoDbQueryableExtensions`
- [ ] Internal `PagingExpression` node (analogous to Cosmos `PagingExpression`) carrying `limit` and `nextToken` parameter names
- [ ] Translation in `DynamoQueryableMethodTranslatingExpressionVisitor` — detect `ToPageAsync`, validate top-level, reject `Limit(n)` combinations; detect `WithNextToken`, store on `SelectExpression`
- [ ] Code generation in `DynamoShapedQueryCompilingExpressionVisitor` — emit `PagingQueryingEnumerable<T>`
- [ ] `PagingQueryingEnumerable<T>` — single-shot executor; resolves params from `QueryContext.Parameters` at runtime; returns `DynamoPage<T>`
- [ ] `SelectExpression` gains nullable `SeedNextToken` field; `QueryingEnumerable` reads it to seed the first request before the auto-page loop
- [ ] Extend `ExecutingExecuteStatement` diagnostic to include `seedNextTokenPresent` bool

### Tests

**Unit (translation):**
- [ ] `ToPageAsync` rejects when not top-level
- [ ] `ToPageAsync` rejects when combined with `Limit(n)`
- [ ] `WithNextToken` rejects when called twice on same query
- [ ] `WithNextToken` stored on `SelectExpression` correctly

**Integration:**
- [ ] `ToPageAsync` first page returns items + non-null `NextToken` when data exceeds limit
- [ ] `ToPageAsync` second page using returned token returns correct items
- [ ] `ToPageAsync` last page returns `NextToken == null`
- [ ] Non-key filter query: `Items.Count < limit` with non-null `NextToken` (demonstrates evaluation-budget semantics)
- [ ] `WithNextToken` + `ToListAsync` auto-collects from cursor onward
- [ ] `WithNextToken` + `Limit(n)` fetches single page from cursor

### Docs

- [ ] New `docs/pagination.md`:
  - Evaluation-budget semantics with concrete example (filtered query returning fewer items than limit with non-null NextToken)
  - Warning callout block (not buried in prose)
  - `ToPageAsync` usage + full example
  - `WithNextToken` usage + full example
  - No backwards pagination, no total count
  - Why `Skip`/`Take` are unsupported
- [ ] XML `<remarks>` on `ToPageAsync` and `WithNextToken` documenting evaluation-budget + non-key filter behavior

## Out of Scope

- Exposing `ConsumedCapacity` on `DynamoPage<T>` (separate issue)
- Deprecating `GetExecuteStatementResponse()` (breaking change, follow-up)
- Changing auto-paging behavior for bare `ToListAsync()`

## Acceptance Criteria

- [ ] `ToPageAsync(25, null)` returns first page with items and non-null `NextToken` when more data exists
- [ ] `ToPageAsync(25, token)` returns next page from correct cursor position
- [ ] Final page returns `NextToken == null`
- [ ] `WithNextToken(token).ToListAsync()` collects all remaining results from that cursor
- [ ] `WithNextToken(token).Limit(25).ToListAsync()` fetches one page from that cursor
- [ ] `ToPageAsync` + `Limit(n)` on same query throws a translation-time error
- [ ] `WithNextToken` called twice on same query throws a translation-time error
- [ ] `DynamoPage<T>` is marked `[Experimental]`
- [ ] `docs/pagination.md` documents evaluation-budget semantics with callout and concrete example

ncipollina:
## Summary
- Add experimental DynamoDB cursor paging APIs: `DynamoPage<T>`, terminal `ToPageAsync(limit, nextToken)`, and composable `WithNextToken(token)`.
- Implement translation/compilation/runtime support for one-request page execution, seeded first-request continuation tokens, and conflict/validation rules aligned with issue #157.
- Extend diagnostics with `seedNextTokenPresent` and add broad unit + integration coverage for paging semantics, token behavior, compiled-query/runtime validation, and re-enumeration behavior.

## Why
- Provide first-class, provider-native cursor paging so callers can request one DynamoDB evaluation-budget page, persist/resume cursors, and avoid ad hoc token plumbing.
- Make paging behavior explicit and predictable for DynamoDB semantics where evaluation budget and returned item count differ.

## Changes
- **Public API**
  - Add `DynamoPage<T>` (`Items`, `NextToken`, `HasMoreResults`) as experimental API.
  - Add `ToPageAsync` and `WithNextToken` extension methods with argument validation and token normalization rules.
- **Query translation / execution**
  - Add `DynamoPagingExpression` and top-level `ToPageAsync` translation path.
  - Reject invalid compositions (e.g. `ToPageAsync` + `Limit`, duplicate `WithNextToken`, unsupported `WithNextToken` with key-only `First*`, ambiguous dual non-null tokens).
  - Add seed token state to `SelectExpression` and carry it into request execution.
  - Add `PagingQueryingEnumerable<T>` for one-request page materialization.
- **Diagnostics**
  - Extend `ExecutingExecuteStatement` logging with `seedNextTokenPresent` (true only for first request seeded by user/query input).
- **Tests**
  - Add/extend unit tests for API validation, translation behavior, runtime token ambiguity handling, and request token propagation.
  - Add/extend integration tests for resume flows, final-page token nulling, single-request seeded limit flow, unseeded multi-request behavior, compiled-query runtime behaviors, and seed reapplication on re-enumeration.
- **Docs**
  - Expand pagination docs with `ToPageAsync`/`WithNextToken` semantics, conflicts, token normalization, diagnostics, and concrete examples.
  - Add pagination link on docs home page.

## Validation
- `dotnet build src/EntityFrameworkCore.DynamoDb/EntityFrameworkCore.DynamoDb.csproj`
- `dotnet test --project tests/EntityFrameworkCore.DynamoDb.IntegrationTests/EntityFrameworkCore.DynamoDb.IntegrationTests.csproj`
- Targeted test classes were iterated during implementation (`LimitTranslationTests`, `PaginationTests`, `CompiledQueryPaginationTests`) and full integration suite is now green.

## Related Issues
- Closes #157


docs/pagination.md
j-d-ha, line 1: This looks good. Can we also mention that you can get the next page token from the response metadata, not just through the page semantics?

src/EntityFrameworkCore.DynamoDb/Query/Internal/DynamoShapedQueryCompilingExpressionVisitor.QueryingEnumerable.cs
```diff
@@ -189,6 +201,186 @@ public ValueTask DisposeAsync()
...
+                    return queryContext.Parameters[parameterExpression.Name] as string;
+
+                throw new InvalidOperationException(
+                    "Seed next-token expression must be normalized before execution.");
+            }
+        }
+    }
+
+#pragma warning disable EF9102
+    private sealed class PagingQueryingEnumerable<T>(
```
j-d-ha, line 190: This class smells a little bit to me, and it feels off that we need to create an entirely new duplicated data structure just to support paging like this. What are your thoughts here? I'm wondering if there's a way to simplify some of this.

src/EntityFrameworkCore.DynamoDb/Query/Internal/DynamoShapedQueryCompilingExpressionVisitor.QueryingEnumerable.cs
```diff
@@ -189,6 +201,186 @@ public ValueTask DisposeAsync()
...
+                    return constantValue;
+
+                if (expression is QueryParameterExpression parameterExpression)
+                    return Convert.ToInt32(queryContext.Parameters[parameterExpression.Name]);
+
+                throw new InvalidOperationException(
+                    "Limit expression must be normalized before execution.");
+            }
+
+            private static string? ResolveStringExpression(
```
j-d-ha: Why do we need this here now? My expectation would be that our existing serialization logic would work here.

tests/EntityFrameworkCore.DynamoDb.IntegrationTests/SimpleTable/PaginationTests.cs
j-d-ha, line 1: Would it not be easier just to assert the generated SQL that gets executed?

j-d-ha: Also, I'm not sure if SqlCapture.Clear(); is actually needed.

j-d-ha: Also, can we add a test that asserts that the data we get back from a page matches the data we get back from regular in-memory data semantics with take?

j-d-ha (commented):
Overall looks really good! I left some questions. GPT also had this to say:

The review output flagged three concrete issues:
1. WithNextToken(...)’s null/whitespace validation is skipped for compiled queries, because the extension method isn’t executed (it’s just an expression node). Translation/execution currently accepts null (and potentially whitespace), which can silently turn “resume from cursor” into “start from beginning”.
2. The ToPageAsync(...) execution path (PagingQueryingEnumerable) doesn’t log ExecutingPartiQlQuery(...), so the PartiQL text is missing from logs for paging.
3. Runtime token extraction uses queryContext.Parameters[name] as string, so wrongly-typed parameter values become null silently, which can mask binding errors (and for WithNextToken, again degrade into “start from beginning”).

docs/pagination.md
ncipollina: Done — added a "NextToken from response metadata" section showing `entry.GetExecuteStatementResponse()?.NextToken` for callers using `ToListAsync`/`Limit` who want the cursor without switching to `ToPageAsync`. Also notes the multi-request case where continuation tokens have already been consumed by the provider.

src/EntityFrameworkCore.DynamoDb/Query/Internal/DynamoShapedQueryCompilingExpressionVisitor.QueryingEnumerable.cs
ncipollina: The structural split is intentional — the two classes produce different sequences (`IAsyncEnumerable<T>` vs `IAsyncEnumerable<DynamoPage<T>>`) with fundamentally different iteration semantics (item-by-item streaming vs single-shot one-page materialization), so a shared `AsyncEnumerator` isn't viable.

What *was* pure duplication was the helper logic. In this commit I extracted `ResolveIntExpression`, `ResolveStringExpression`, and `NormalizeToken` as `private static` methods on the outer `DynamoShapedQueryCompilingExpressionVisitor` class — both enumerables share them now. Also wired the missing `ExecutingPartiQlQuery` log call into `PagingQueryingEnumerable` while in there.

src/EntityFrameworkCore.DynamoDb/Query/Internal/DynamoShapedQueryCompilingExpressionVisitor.QueryingEnumerable.cs
ncipollina: Compiled queries never execute `WithNextToken(token)` directly — the argument is captured as a `QueryParameterExpression` node in the expression tree rather than a `ConstantExpression`. At execution time the actual token value lives in `DynamoQueryContext.Parameters[name]`, so the lookup is required or the token is always null (silent start-from-beginning for compiled-query callers).

The existing serialization logic handles constants fine; this fills the compiled-query gap. The method is now extracted to the outer class and shared between both enumerables, with the rationale in its XML comment.

tests/EntityFrameworkCore.DynamoDb.IntegrationTests/SimpleTable/PaginationTests.cs
ncipollina: `AssertSql` checks PartiQL text only. The interesting pagination behavior — `Limit` on the request, `NextToken` presence flags — is not in the SQL; it's on `ExecuteStatementRequest`. `ExecuteStatementCalls` is the right tool for asserting those semantics, so the approach stays.

Also addressed the other two points in this thread:
- Removed all redundant `SqlCapture.Clear()` calls at the top of test methods — the `DynamoTestFixtureBase` constructor already clears on each xUnit test instance. Mid-test clears (isolating the second operation in two-leg tests) are preserved.
- Added `ToPageAsync_Items_MatchInMemoryTakeSemantics`: fetches all seeded items in one page and asserts order-insensitive equivalence to `SimpleItems.Items`.

j-d-ha (approved):
🚀 Awesome work!

PR #162, opened 2026-04-16
fix(dynamodb): align key mapping and naming behavior for table keys

j-d-ha:
## Summary
- Apply attribute naming conventions to user-defined shadow properties while still skipping provider-internal runtime-only and owned-ordinal metadata properties.
- Make `HasPartitionKey(...)` and `HasSortKey(...)` lambda overloads map the selected member up front, and fail model validation early when configured table keys are shadow/runtime-only properties.
- Update unit/integration coverage and docs to clarify mapped-property key discovery, key materialization requirements, and supported key modeling patterns.

## Validation
- `tests/EntityFrameworkCore.DynamoDb.Tests/EntityFrameworkCore.DynamoDb.Tests.csproj` (366 passed)
- `tests/EntityFrameworkCore.DynamoDb.IntegrationTests/EntityFrameworkCore.DynamoDb.IntegrationTests.csproj` (331 passed, 1 skipped)

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -240,6 +240,31 @@ b.HasPartitionKey("CustomerId");
 b.HasSortKey("OrderId");
 ```
 
+### Key property requirements
+
+Partition/sort key properties must be mapped as real EF properties that can be materialized.
+
+- Supported: normal settable properties, private-set properties, or getter-only properties with a
+    mapped backing field.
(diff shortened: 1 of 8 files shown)


ncipollina (approved):
✅ I'm not fully following what changed here, but looks fine to me.

PR #164, opened 2026-04-17
fix(shared-table): respect HasNoDiscriminator for shared-table mappings

### LINKED ISSUE #163: [Bug]: HasNoDiscriminator() is ignored for shared-table mappings and '' is still injected
## What happened?

When multiple concrete entity types are mapped to the same DynamoDB table, the provider conventionally adds a shadow discriminator attribute named `$type`. Calling EF Core's `HasNoDiscriminator()` should disable discriminator usage for that shared-table mapping, but the provider still behaves as if a discriminator is mandatory.

Expected behavior:
- `HasNoDiscriminator()` should prevent the provider from appending the discriminator shadow property/attribute to requests for that shared table.
- Queries against that table should not automatically inject `$type` predicates or projections.
- Shared-table mappings that intentionally do not persist a discriminator field should be allowed to run.

Actual behavior:
- `$type` is still appended/required for shared-table requests.
- Query translation still injects discriminator filtering whenever more than one concrete type shares a table.
- Materialization/validation still treats a discriminator as required for the shared table.

While reviewing the code on `main` at `cf58bb4`, I found several paths that currently enforce this behavior:
- `src/EntityFrameworkCore.DynamoDb/Metadata/Conventions/DynamoDiscriminatorConvention.cs` unconditionally re-adds a discriminator for any shared table group with more than one concrete entity type (`ProcessModelFinalizing`, lines 31-55).
- `src/EntityFrameworkCore.DynamoDb/Query/Internal/DynamoQueryableMethodTranslatingExpressionVisitor.cs` injects discriminator predicates whenever a table group contains more than one concrete type (`CreateDiscriminatorPredicate` / `RequiresDiscriminatorPredicate`, lines 426-500).
- `src/EntityFrameworkCore.DynamoDb/Query/Internal/Expressions/SelectExpression.cs` adds the discriminator shadow property to projections when present (`AddDiscriminatorToProjection`, lines 331-340).
- `src/EntityFrameworkCore.DynamoDb/Infrastructure/Internal/DynamoModelValidator.cs` requires a discriminator property and value for shared-table mappings with multiple concrete types (`ValidateDiscriminatorMappingGroup`, lines 486-560).

This makes `HasNoDiscriminator()` ineffective for the single-table/shared-table scenario.

## Steps to reproduce

1. Configure two concrete entity types to map to the same DynamoDB table with the provider's shared-table pattern.
2. Configure the shared-table mapping to opt out of EF Core discriminator metadata with `HasNoDiscriminator()`.
3. Use a table/items that do not contain a persisted `$type` attribute.
4. Execute a query or save operation against one of those entity sets.
5. Observe that the generated request still includes `$type` behavior (predicate/projection/write value), or that model validation/materialization still requires the discriminator.

A minimal example shape is:

```csharp
modelBuilder.Entity<UserEntity>(b =>
{
    b.ToTable("app-table");
    b.HasPartitionKey(x => x.Pk);
    b.HasSortKey(x => x.Sk);
    b.HasNoDiscriminator();
});

modelBuilder.Entity<OrderEntity>(b =>
{
    b.ToTable("app-table");
    b.HasPartitionKey(x => x.Pk);
    b.HasSortKey(x => x.Sk);
});
```

## Environment

- Repository: `LayeredCraft/dynamodb-efcore-provider`
- Revision reviewed: `cf58bb4` (`main`)
- OS: `Darwin 25.4.0`
- .NET SDK: `10.0.103`
- Target framework in repo: `net10.0`

j-d-ha:
## Summary
Fix shared-table mappings so `HasNoDiscriminator()` is honored (no `$type` projection/predicate/validation) by moving discriminator handling to a single event-driven convention (Cosmos-style) rather than relying on model-finalizing only.

## Changes
- Replace EF Core `DiscriminatorConvention` with a unified `DynamoDiscriminatorConvention` that reacts to entity/table/base-type/discriminator-name changes and applies table-group decisions incrementally.
- Keep a small model-finalizing cleanup to remove convention-only discriminators for final single-type, non-hierarchy groups.
- Allow shared-table groups with discriminator disabled in `DynamoModelValidator` (and throw on mixed configuration).
- Add unit/integration coverage for `HasNoDiscriminator()` (including call-order regression) and query baselines without `$type`.

## Validation
- `dotnet test` (solution)
- Unit tests: `tests/EntityFrameworkCore.DynamoDb.Tests`
- Integration tests: `tests/EntityFrameworkCore.DynamoDb.IntegrationTests`

## Related Issues
Fixes #163


src/EntityFrameworkCore.DynamoDb/Infrastructure/Internal/DynamoModelValidator.cs
```diff
@@ -505,7 +505,29 @@ private static void ValidateDiscriminatorMappingGroup(
...
+        if (concreteTypesWithDiscriminator.Count == 0)
+        {
+            if (rootEntityTypes.All(IsDiscriminatorDisabled))
+                return;
+
+            throw new InvalidOperationException(
+                $"Entity types mapped to shared DynamoDB table '{tableName}' do not define a discriminator property. "
+                + "Shared-table mappings with multiple entity types require a discriminator by default. "
+                + "Call HasNoDiscriminator() for the table group to explicitly opt out.");
+        }
```
ncipollina, line 522: ❓ Does this mean that we are requiring an explicit discriminator to be defined?

ncipollina (commented):
One question, otherwise looks good.

src/EntityFrameworkCore.DynamoDb/Infrastructure/Internal/DynamoModelValidator.cs
j-d-ha: No, we default to having a discriminator now and if you don't want it you need to explicitly say so.

ncipollina (dismissed):
I still approve.

PR #181, opened 2026-04-22
fix(storage): omit null GSI key attributes from PartiQL INSERT

### LINKED ISSUE #180: [Bug]: Null GSI key attributes cause AmazonDynamoDBException on SaveChanges INSERT
## What happened?

When an entity has nullable GSI key properties (sparse GSI pattern) and those properties are `null`, `SaveChangesAsync` throws `Amazon.DynamoDBv2.AmazonDynamoDBException: Invalid attribute value type`.

**Expected:** Items with null GSI key properties should be saved successfully. DynamoDB's sparse GSI pattern works by simply omitting attributes from items that should not appear in the index.

**Actual:** The provider serializes null strings as `{ NULL: true }` AttributeValue, then includes those in the PartiQL `INSERT INTO ... VALUE { ... }` statement. DynamoDB rejects `{ NULL: true }` as a value for any attribute that is declared as a GSI key attribute in the table definition.

## Steps to reproduce

1. Define a table with a GSI (e.g., `gs1-pk` / `gs1-sk`).
2. Map an EF entity with nullable GSI key properties configured via `HasGlobalSecondaryIndex`.
3. Insert an entity where the GSI key properties are `null`.
4. Call `SaveChangesAsync`.
5. Observe: `AmazonDynamoDBException: Invalid attribute value type`.

Minimal model:
```csharp
public sealed class SparseItem
{
    public string Pk { get; set; } = null!;
    public string Sk { get; set; } = null!;
    public string? Gs1Pk { get; set; }  // null → not in the GSI
    public string? Gs1Sk { get; set; }
}
```

Model config:
```csharp
builder.HasGlobalSecondaryIndex("gs1-index", nameof(SparseItem.Gs1Pk), nameof(SparseItem.Gs1Sk));
```

## Environment

.NET 10, EntityFrameworkCore.DynamoDb 0.0.2-alpha.6, DynamoDB Local (Testcontainers) and AWS DynamoDB.

ncipollina:
## Summary

DynamoDB rejects `{ NULL: true }` for attributes that participate in a GSI key definition. When an entity uses a sparse GSI pattern (nullable GSI key properties that are `null` for items that should not appear in the index), `SaveChangesAsync` threw `AmazonDynamoDBException: Invalid attribute value type` because the provider was including those properties as `NULL`-typed AttributeValues in the PartiQL `INSERT INTO ... VALUE { ... }` statement.

`DynamoSaveChangesPlanner` now strips any GSI key attribute with a `{ NULL: true }` value from the item before building the INSERT statement, matching DynamoDB's natural sparse-GSI behaviour (absent attributes = item not indexed). Regular nullable scalar properties are unaffected and continue to write as `{ NULL: true }`.

## Changes

**Fix — `DynamoSaveChangesPlanner`**
- On `EntityState.Added`, iterate the entity type's indexes; for each GSI, remove any entry from the item where `value.NULL == true` before passing to `BuildInsertStatement`.
- `BuildInsertStatement` itself is unchanged — the fix is scoped to INSERT, where absent = sparse; UPDATE uses SET/REMOVE which already handles this correctly.

**Tests — `SaveChangesTable`**
- `Table.cs`: added `gs1-pk` / `gs1-sk` attribute definitions and a `gs1-index` GSI to the shared `AppItems` table.
- `Models.cs`: added `SparseGsiItem` — a record with nullable `Gs1Pk` / `Gs1Sk` properties.
- `DbContext.cs`: registered `SparseGsiItem` with `HasGlobalSecondaryIndex` configuration.
- `SparseGsiSaveChangesTests.cs`: three new integration tests:
  - `AddAsync_WithNullGsiKeys_GsiKeyAttributesAbsentFromItem` — verifies no exception and GSI attributes absent from raw item.
  - `AddAsync_WithNullGsiKeys_NullableScalar_IsWrittenAsNullType` — regression guard: non-GSI nullable scalars still write as `{ NULL: true }`.
  - `AddAsync_WithPopulatedGsiKeys_WritesGsiKeyAttributesToItem` — verifies GSI attributes present with correct values when populated.

## Validation

- `dotnet build` on provider and integration test project: 0 errors, 0 warnings.
- Integration tests exercised against DynamoDB Local via Testcontainers.

## Notes for Reviewers

The fix intentionally targets only the `EntityState.Added` path in `DynamoSaveChangesPlanner`. The `Modified` path uses PartiQL `UPDATE ... SET / REMOVE`, where `REMOVE` already handles absent-attribute semantics correctly for null GSI keys — no change needed there.

An earlier attempt applied the null-skip inside `BuildInsertStatement` unconditionally, which broke the existing `AddAsync_NullableScalar_IsWrittenAsNull` test (non-GSI nullable scalars must still write as `{ NULL: true }`). The per-GSI-key targeted approach fixes that regression.

## Related Issues

Closes #180

### DIFF
--- a/src/EntityFrameworkCore.DynamoDb/Storage/DynamoSaveChangesPlanner.cs
+++ b/src/EntityFrameworkCore.DynamoDb/Storage/DynamoSaveChangesPlanner.cs
@@ -49,6 +49,21 @@ internal sealed class DynamoSaveChangesPlanner(
                 {
                     var item = serializerSource.BuildItem(entry);
                     var tableName = (string)entry.EntityType[DynamoAnnotationNames.TableName]!;
+
+                    // DynamoDB rejects { NULL: true } for GSI key attributes via PartiQL INSERT.
+                    // Sparse GSIs require these attributes to simply be absent when not applicable.
+                    foreach (var index in entry.EntityType.GetIndexes())
+                    {
+                        if (index.GetSecondaryIndexKind() is null)
+                            continue;
+                        foreach (var property in index.Properties)
+                        {
+                            var attrName = property.GetAttributeName();
+                            if (item.TryGetValue(attrName, out var val) && val.NULL == true)
+                                item.Remove(attrName);
+                        }
+                    }
+
                     var (sql, parameters) = statementFactory.BuildInsertStatement(tableName, item);
 
                     AddCompiledOperation(
--- a/.github/workflows/pr-build.yaml
+++ b/.github/workflows/pr-build.yaml
@@ -7,7 +7,7 @@ on:
 permissions: write-all
 jobs:
   build:
-    uses: LayeredCraft/devops-templates/.github/workflows/pr-build.yaml@v8.0
+    uses: LayeredCraft/devops-templates/.github/workflows/pr-build.yaml@v8.1
     with:
       solution: EntityFrameworkCore.DynamoDb.slnx
       hasTests: true
--- a/.github/workflows/pr-title-check.yaml
+++ b/.github/workflows/pr-title-check.yaml
@@ -10,4 +10,4 @@ permissions:
 
 jobs:
   validate:
-    uses: LayeredCraft/devops-templates/.github/workflows/pr-title-check.yml@v8.0
+    uses: LayeredCraft/devops-templates/.github/workflows/pr-title-check.yml@v8.1
--- a/.github/workflows/publish-preview.yaml
+++ b/.github/workflows/publish-preview.yaml
@@ -19,7 +19,7 @@ jobs:
   # (e.g. release/net9) and wire up separate caller workflows targeting those branches.
 
   publish:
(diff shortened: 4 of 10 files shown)


src/EntityFrameworkCore.DynamoDb/Storage/DynamoSaveChangesPlanner.cs
```diff
@@ -49,6 +49,21 @@ private List<CompiledWriteOperation> BuildWriteOperations(
...
+                        if (index.GetSecondaryIndexKind() is null)
+                            continue;
+                        foreach (var property in index.Properties)
+                        {
+                            var attrName = property.GetAttributeName();
+                            if (item.TryGetValue(attrName, out var val) && val.NULL == true)
+                                item.Remove(attrName);
+                        }
+                    }
+
```
j-d-ha, line 66: Nice fix. We can definitely extend this later once we add an option in for other values.

j-d-ha (approved):
Great work!

PR #185, opened 2026-04-23
fix(conventions): remove Id-based key discovery for owned entity types

### LINKED ISSUE #184: Remove Id-based key discovery convention for owned entity types
## Summary

The DynamoDB provider inherits EF Core's standard `KeyDiscoveryConvention` for owned entity types (via `DynamoKeyDiscoveryConvention.ProcessKeyProperties` delegating to `base` when `entityType.IsOwned()`). This means any owned type with a CLR property named `Id` or `<TypeName>Id` will have that property added to its EF primary key, which for owned collection types prevents `OwnedTypePrimaryKeyConvention` from adding the required `__OwnedOrdinal` shadow key during model finalization.

## Reproduction

```csharp
public record DuplicateResult
{
    public required string Id { get; set; }  // triggers KeyDiscoveryConvention
    public required float Score { get; set; }
}

// In OnModelCreating:
builder.OwnsMany(x => x.DuplicatesFound, duplicate =>
{
    duplicate.Property(d => d.Score).HasConversion(...);
});
```

This throws at runtime:

```
System.InvalidOperationException: Owned collection element entity type 'DuplicateResult'
does not have an ordinal key property. Owned collection elements require a synthetic
ordinal key for stable identity and change tracking.
```

## Root Cause

EF Core's `KeyDiscoveryConvention` treats a property named `Id` as a primary key candidate for owned entity types. When the `OwnsMany` lambda performs any explicit configuration, `Id` gets set in the owned type's PK at explicit configuration source. `OwnedTypePrimaryKeyConvention` (a finalizing convention, which runs at convention source) cannot override an explicitly-set PK, so `__OwnedOrdinal` is never successfully added.

DynamoDB has no concept of an `Id`-based key convention — partition/sort keys are the only meaningful table keys. For owned types specifically:
- **OwnsOne**: key is set by EF Core's FK conventions pointing to the owner's PK
- **OwnsMany**: key is set entirely by `OwnedTypePrimaryKeyConvention` (FK + `__OwnedOrdinal`)

Neither path has any use for `Id`-based key discovery. Running the base `KeyDiscoveryConvention` for owned types is not only unnecessary but actively harmful.

## Expected Behavior

`OwnedTypePrimaryKeyConvention` should always be able to add `__OwnedOrdinal` as the ordinal key for owned collection element types, regardless of whether the CLR type has a property named `Id`.

## Suggested Fix

The simplest fix is to stop calling `base.ProcessKeyProperties` for owned types entirely in `DynamoKeyDiscoveryConvention`. Owned type keys are managed by EF Core's FK conventions and `OwnedTypePrimaryKeyConvention` — not by name-based discovery.

```csharp
protected override void ProcessKeyProperties(
    IList<IConventionProperty> keyProperties,
    IConventionEntityType entityType)
{
    if (entityType.IsOwned())
    {
        // DynamoDB has no Id-based key convention. Owned type keys are managed
        // by EF FK conventions (OwnsOne) and OwnedTypePrimaryKeyConvention (OwnsMany).
        return;
    }

    // existing root entity logic...
}
```

This is preferable to a narrower fix (e.g. only suppressing for non-unique owned types) because `Id` discovery is meaningless for all owned types in DynamoDB, not just collection elements.

## Workaround

Rename any `Id` property on owned collection element types to something that does not match EF Core's `Id`/`<TypeName>Id` naming convention, and use `HasAttributeName("id")` in the `OwnsMany` configuration to preserve the DynamoDB attribute name.

ncipollina:
## Summary

- Stop delegating to `base.ProcessKeyProperties` for owned entity types in `DynamoKeyDiscoveryConvention`
- Clears key candidates and returns early for all owned types instead
- Prevents EF Core's `Id`/`<TypeName>Id` naming heuristic from adding a property to the owned entity's PK at explicit config source, which was blocking `OwnedTypePrimaryKeyConvention` from adding the required `__OwnedOrdinal` shadow key

## Test plan

- [ ] Existing unit tests pass
- [ ] Existing integration tests pass
- [ ] Owned collection element type with a property named `Id` no longer throws `InvalidOperationException` at model validation

Closes #184

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/EntityFrameworkCore.DynamoDb/Metadata/Conventions/DynamoKeyDiscoveryConvention.cs
+++ b/src/EntityFrameworkCore.DynamoDb/Metadata/Conventions/DynamoKeyDiscoveryConvention.cs
@@ -46,9 +46,24 @@ public class DynamoKeyDiscoveryConvention(ProviderConventionSetBuilderDependenci
     {
         if (entityType.IsOwned())
         {
-            // DynamoDB has no Id-based key convention. Owned type keys are managed
-            // by EF FK conventions (OwnsOne) and OwnedTypePrimaryKeyConvention (OwnsMany).
-            keyProperties.Clear();
+            var ownership = entityType.FindOwnership();
+            if (ownership is { IsUnique: false })
+            {
+                // For owned collection elements (OwnsMany), let base deduplicate the
+                // pre-populated [FK..., shadow_ordinal] list, then strip any non-FK,
+                // non-int candidates (e.g. a CLR 'Id' property discovered by EF naming
+                // conventions). DynamoDB owned collection PKs are FK props + int ordinal only.
+                base.ProcessKeyProperties(keyProperties, entityType);
+                var fkSet = new HashSet<IConventionProperty>(ownership.Properties);
+                for (var i = keyProperties.Count - 1; i >= 0; i--)
+                {
+                    if (!fkSet.Contains(keyProperties[i]) && keyProperties[i].ClrType != typeof(int))
+                        keyProperties.RemoveAt(i);
+                }
+                return;
+            }
+
+            base.ProcessKeyProperties(keyProperties, entityType);
             return;
         }
 
--- a/src/EntityFrameworkCore.DynamoDb/Metadata/Conventions/DynamoKeyDiscoveryConvention.cs
+++ b/src/EntityFrameworkCore.DynamoDb/Metadata/Conventions/DynamoKeyDiscoveryConvention.cs
@@ -27,10 +27,70 @@ namespace EntityFrameworkCore.DynamoDb.Metadata.Conventions;
 ///     conventional partition key name do not fall back to EF Core's <c>Id</c>/<c>[Key]</c> discovery.
 ///     Annotation-setting and ambiguity validation are handled separately by
 ///     <c>DynamoKeyAnnotationConvention</c>.
+///     <para>
+///         For owned collection element types (<c>OwnsMany</c>), EF Core 10's base
+///         <c>DiscoverKeyProperties</c> calls <c>CreateUniqueProperty(int, "Id")</c> when no
+///         naming-convention candidates are found, producing a shadow int property named "Id".
+///         If the user later explicitly configures a CLR property also named "Id" (e.g.,
+///         <c>results.Property(r =&gt; r.Id).HasAttributeName("id")</c>), the shadow and CLR
+///         properties conflict, triggering repeated convention invocations until EF Core hits
+///         its recursion limit. This convention overrides <c>DiscoverKeyProperties</c> for
+///         <c>OwnsMany</c> types to bypass that path entirely; ordinal shadow key creation
+///         is left to <c>OwnedTypePrimaryKeyConvention</c> during model finalization.
+///     </para>
 /// </remarks>
 public class DynamoKeyDiscoveryConvention(ProviderConventionSetBuilderDependencies dependencies)
     : KeyDiscoveryConvention(dependencies)
 {
+    /// <summary>
+    ///     Returns the candidate key properties for the entity type, bypassing EF Core's
+    ///     naming-convention discovery for owned collection elements.
+    /// </summary>
+    /// <remarks>
+    ///     For <c>OwnsMany</c> entity types, returns the ownership FK properties plus an ordinal
+    ///     shadow int property. If no ordinal exists yet, one is created immediately using the
+    ///     <c>__OwnedOrdinal</c> base name via <c>CreateUniqueProperty</c>. Creating the ordinal
+    ///     here (during model building) rather than in <c>OwnedTypePrimaryKeyConvention</c>
+    ///     (finalization) avoids exceeding EF Core's convention-invocation recursion limit.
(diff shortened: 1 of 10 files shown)


ncipollina:
@codex review

ncipollina:
@codex review

tests/EntityFrameworkCore.DynamoDb.IntegrationTests/OwnedTypesTable/OwnedCollectionWithIdPropertyTests.cs
```diff
@@ -0,0 +1,131 @@
...
+        using var readCtx = OwnedCollectionWithIdPropertyDbContext.Create(Client);
+        var loaded = await readCtx.Reports
+            .Where(r => r.Pk == "ANALYSIS#EMPTY1")
+            .AsAsyncEnumerable()
+            .SingleAsync(CancellationToken);
+
+        loaded.Pk.Should().Be("ANALYSIS#EMPTY1");
+        loaded.Results.Should().BeEmpty();
+    }
+}
```
j-d-ha, line 131: Missing trailing newline at end of file. Most editors/linters (and `git diff`) will flag this.

tests/EntityFrameworkCore.DynamoDb.IntegrationTests/OwnedTypesTable/Infra/TestFixture.cs
```diff
@@ -3,6 +3,25 @@
...
+        => EnsureClassTableInitialized(
+            AnalysisReportTable.TableName,
+            AnalysisReportTable.CreateTable);
+
+    public OwnedCollectionWithIdPropertyDbContext Db
+    {
+        get
+        {
+            field ??= new OwnedCollectionWithIdPropertyDbContext(
+                CreateOptions<OwnedCollectionWithIdPropertyDbContext>(
```
j-d-ha, line 18: The `field` keyword (auto-implemented property backing field) is a C# 13 feature. If the project targets .NET 8 / C# 12, this will fail to compile. Please confirm `<LangVersion>` allows it, or replace with a conventional backing field:

```csharp
private OwnedCollectionWithIdPropertyDbContext? _db;
public OwnedCollectionWithIdPropertyDbContext Db
    => _db ??= new OwnedCollectionWithIdPropertyDbContext(
        CreateOptions<OwnedCollectionWithIdPropertyDbContext>(
            options => options.DynamoDbClient(Client)));
```

src/EntityFrameworkCore.DynamoDb/Metadata/Conventions/DynamoKeyDiscoveryConvention.cs
```diff
@@ -27,10 +28,70 @@ namespace EntityFrameworkCore.DynamoDb.Metadata.Conventions;
...
+                // without recursing further. Creating it here (during model building, not
+                // finalization) keeps total convention invocations within EF Core's recursion limit.
+                var ordinalBuilder = entityType.Builder.CreateUniqueProperty(typeof(int), "__OwnedOrdinal", true);
+                if (ordinalBuilder != null)
+                {
+                    ordinalBuilder.HasAnnotation(DynamoAnnotationNames.OwnedOrdinalKey, true);
+                    keyProps.Add(ordinalBuilder.Metadata);
+                }
+                return keyProps;
+            }
```
j-d-ha, line 93: If `CreateUniqueProperty` returns `null` (builder locked during convention re-entry), the method returns FK-only key candidates with no ordinal. EF Core would then call `ProcessKeyProperties` with an incomplete key, and `OwnedTypePrimaryKeyConvention` may or may not recover at finalization. Consider throwing here so the failure is visible rather than silently producing a broken PK:

```csharp
if (ordinalBuilder == null)
    throw new InvalidOperationException(
        $"Unable to create ordinal shadow property on owned type '{entityType.DisplayName()}'. " +
        "The model builder may be locked.");
```

j-d-ha (commented):
A few things that need fixing before merge — all other feedback is informational only.

tests/EntityFrameworkCore.DynamoDb.IntegrationTests/OwnedTypesTable/OwnedCollectionWithIdPropertyTests.cs
ncipollina: Minor issue, not really worth dealing with.

tests/EntityFrameworkCore.DynamoDb.IntegrationTests/OwnedTypesTable/Infra/TestFixture.cs
ncipollina: Our package will be compiled so not an issue. Not to mention, this is our unit test.

src/EntityFrameworkCore.DynamoDb/Metadata/Conventions/DynamoKeyDiscoveryConvention.cs
ncipollina: Fixed.

j-d-ha (approved):
🤩 awesome fix!

PR #187, opened 2026-04-24
fix(query): omit Parameters from ExecuteStatementRequest when empty

### LINKED ISSUE #186: [Bug]: ExecuteStatementRequest sends empty Parameters list causing DynamoDB validation error
## What happened?

When executing a PartiQL `SELECT` query that contains **no parameterized placeholders** (e.g., all values are inlined as literals), the provider constructs an `ExecuteStatementRequest` with `Parameters = []` (an empty list). DynamoDB's `ExecuteStatement` API rejects this with a validation error:

```
Amazon.DynamoDBv2.AmazonDynamoDBException: 1 validation error detected:
Value '[]' at 'parameters' failed to satisfy constraint:
Member must have length greater than or equal to 1
```

**Expected behaviour:** When the PartiQL query has no parameterized values, the `Parameters` field should be omitted (`null`) rather than set to an empty list. DynamoDB requires the field to either be absent or contain at least one entry.

**Root cause:** In `DynamoShapedQueryCompilingExpressionVisitor.QueryingEnumerable.cs`, the `ExecuteStatementRequest` is constructed with:

```csharp
Parameters = sqlQuery.Parameters.ToList(),
```

This always sets `Parameters`, even when `sqlQuery.Parameters` is empty. The fix is to only set the field when parameters are present:

```csharp
Parameters = sqlQuery.Parameters.Count > 0 ? sqlQuery.Parameters.ToList() : null,
```

This occurs in two places in the file (the paginated enumerable path and the single-page `ToList` path).

## Steps to reproduce

1. Configure a `DbContext` with a DynamoDB entity that has a GSI (Global Secondary Index).
2. Execute a query that filters on GSI partition key and a discriminator (`$type`) — both values are emitted as string literals in the PartiQL, not as `?` placeholders.
3. Example query: `SELECT * FROM "table-name"."gsi-index" WHERE "gs1-pk" = 'game' AND "$type" = 'TriviaGame'`
4. Observe the `AmazonDynamoDBException` at runtime.

## Environment

- .NET 10
- `EntityFrameworkCore.DynamoDb` 0.1.0-alpha.18
- AWS SDK for .NET (`AWSSDK.DynamoDBv2`)
- macOS / AWS Lambda (reproducible in both environments)

ncipollina:
## Summary

- DynamoDB rejects `ExecuteStatement` calls where `Parameters = []` (empty list) — it requires the field to be absent or contain ≥ 1 entry
- When PartiQL queries use only inlined literal values (no `?` placeholders), `sqlQuery.Parameters` is empty and was unconditionally sent as an empty list
- Fixed by guarding with `Count > 0` before setting `Parameters`, falling back to `null` — applies to both the paginated enumerable path and the single-page `ToList` path in `QueryingEnumerable.cs`, and the write path in `DynamoClientWrapper.cs`

## Test plan

- [ ] Existing tests pass
- [ ] Execute a GSI query where all predicate values are inlined literals (e.g. discriminator + partition key constant) — no `AmazonDynamoDBException` with `parameters` constraint violation

Closes #186

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/EntityFrameworkCore.DynamoDb/Query/Internal/DynamoShapedQueryCompilingExpressionVisitor.QueryingEnumerable.cs
+++ b/src/EntityFrameworkCore.DynamoDb/Query/Internal/DynamoShapedQueryCompilingExpressionVisitor.QueryingEnumerable.cs
@@ -146,7 +146,7 @@ public partial class DynamoShapedQueryCompilingExpressionVisitor
                         new ExecuteStatementRequest
                         {
                             Statement = sqlQuery.Sql,
-                            Parameters = sqlQuery.Parameters.ToList(),
+                            Parameters = sqlQuery.Parameters.Count > 0 ? sqlQuery.Parameters.ToList() : null,
                             // Maps directly to ExecuteStatementRequest.Limit (evaluation budget).
                             Limit = _limit,
                             NextToken = _seedNextToken,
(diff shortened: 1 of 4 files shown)


j-d-ha (approved):
Awesome work!

PR #190, opened 2026-04-30
feat(complex-types): replace owned entity support with complex properties

j-d-ha:
## Summary
This PR completes the provider migration from owned-entity modeling to EF Core complex properties so model configuration, query translation, and SaveChanges behavior align with the complex-type architecture. It removes owned-type-only infrastructure and updates docs for the new modeling surface, including clarifications for supported CLR types in complex collections.

## Changes
- Replaced owned-type metadata/convention paths with complex-property discovery, containment, and validation conventions.
- Updated query translation and projection/materialization to operate on complex-property shapes, including nested and collection paths.
- Refactored SaveChanges serialization/planning to write complex-property graphs without owned-type infrastructure.
- Removed obsolete owned artifacts (conventions, annotations, helpers, and owned-only tests).
- Added/migrated unit and integration coverage for discovery, complex collection shape handling, key/query behavior, and save pipeline correctness.
- Refreshed provider docs for modeling, limitations, querying, diagnostics, and attribute naming with explicit guidance for complex collections.

## Validation
- Expanded and migrated tests in `tests/EntityFrameworkCore.DynamoDb.Tests` and `tests/EntityFrameworkCore.DynamoDb.IntegrationTests` to cover complex-type discovery, translation, serialization, and execution behavior.
- Added timeout-backed facts via `TestConfiguration.DefaultTimeout` to improve reliability and avoid hanging runs.


ncipollina:
@codex review

ncipollina (commented):
## PR Review: feat(complex-types): replace owned entity support with complex properties

Well-structured replacement of a whole modelling subsystem. Core logic is sound, new expression nodes are clean, cycle-detection algorithm is correct. No critical bugs found. Findings below are primarily doc/polish issues with two minor correctness/clarity items worth addressing before merge.

---

### Critical

None.

---

### Major

**1. Stale `#pragma warning disable EF1001` in `DynamoSaveChangesPlanner.cs` (line 8)**

The suppress comment reads `// Internal EF Core API: InternalEntityEntry.SetProperty`, but `InternalEntityEntry` is no longer referenced anywhere in the file body after the refactor. The pragma suppresses `EF1001` for code that no longer exists, which violates the project rule that suppressions must be per-file with a short reason comment describing what is actually suppressed. Either remove the pragma entirely if no EF1001 warning is generated, or update the comment to name the actual internal API in use.

**2. Dead private methods in `DynamoProjectionBindingRemovingExpressionVisitor` (lines ~887–1010)**

Four private static methods — `IsWirePrimitiveType`, `CreateAttributeValueToPrimitiveExpression`, `GetExpectedWireMemberName`, and `TryCreateHasWireValueExpression` — are declared in the class but have zero call sites anywhere in the codebase. These are almost certainly remnants of the old owned-entity deserialization path. Dead code in a query-compilation hot path is maintenance debt: readers cannot know whether these methods were intentionally preserved for a future feature or are safe to delete. They should be removed, or — if deliberately scaffolding for planned work — a comment should explain that.

---

### Minor

**3. Broken `<param>` XML documentation in `DynamoEntityTypeBuilderExtensions.cs`**

Several method parameter descriptions are written as free-running `///` lines instead of proper `<param name="...">` tags. Affected sites include `ToTable(string? name)` (both overloads), `HasPartitionKey(string propertyName)`, and both `HasPartitionKey`/`HasSortKey` on the generic overload. These loose lines are invisible to tooling and violate the CLAUDE.md requirement that public methods include `<param>` where parameter names are not self-explanatory. Note also that `null` literals are missing — the text reads `Pass  to clear` with empty slots where `<see langword="null" />` should be.

**4. Stale "owned entity types" references in `HasAttributeNamingConvention` remarks**

`DynamoEntityTypeBuilderExtensions.cs` — lines ~133–136, ~159–162, ~299–302, ~322–325

The `<remarks>` blocks for both `HasAttributeNamingConvention` overloads still read:
> Owned entity types without their own convention configured inherit this entity's convention automatically.

and

> runtime-only and owned-ordinal shadow properties are not affected.

Since this PR removes owned entity support entirely, these remarks are now incorrect. Replace with the equivalent statements about complex types, and drop the "owned-ordinal shadow properties" reference.

**5. Missing `<summary>` on `BuildModifiedUpdateStatement`, `BuildDeleteStatement`, and `BuildInsertStatement` in `DynamoPartiqlStatementFactory.cs`**

The three `internal` top-level builder methods have no XML documentation, even though CLAUDE.md requires `<summary>` on all methods. The private helpers added in this PR (`AppendComplexPropertyMutations`, `AppendComplexCollectionMutations`, `ShouldReplaceWholeComplexProperty`) all have summaries; the three entry points should too.

**6. `DynamoProjectionBindingExpressionVisitor.Translate` does not clean up on exception**

`DynamoProjectionBindingExpressionVisitor.cs` — lines ~29–66

The cleanup block at the end of `Translate` (resetting `_projectionMapping`, `_selectExpression`, `_projectionMembers`, `_indexBasedBinding`) runs only on the happy path. If any `Visit*` method throws, the visitor is left in a partially-mutated state. `DynamoSqlTranslatingExpressionVisitor.Translate` guards against this correctly with a `try/finally` block. The projection binding visitor should do the same.

---

### Suggestions

**7. `DynamoComplexPropertyDiscoveryConvention` is `public sealed` — consider `internal`**

The class overrides an EF Core internal convention and is registered only via `DynamoConventionSetBuilder`. Making it `public` widens the surface area. `DynamoComplexContainmentValidationConvention` and `DynamoOwnedEntityTypeValidationConvention` are also `public sealed` — worth confirming all three need to be public vs. `internal sealed` for consistency and to minimize the exposed API surface.

**8. `ShouldReplaceWholeComplexProperty` nested-collection branch does not short-circuit symmetrically**

`DynamoPartiqlStatementFactory.cs` — lines ~349–356

A modified nested collection sets `hasModifiedMember = true` but doesn't return early, while the scalar branch returns `false` early. Functionally correct, but the asymmetry is harder to read. A comment explaining why the collection branch doesn't return early (or restructuring for consistency) would help future readers.

**9. `DynamoComplexContainmentValidator.CreateCycleException` — clarify `chain[0]` intent**

`DynamoComplexContainmentValidator.cs` — lines ~59–63

When a cycle is detected, `chain[0]` is always the first complex property on the *current entity type being validated* (since `chain` resets per entity). This is correct and intentional, but a short comment would clarify that the root in the error message is the root of the current DFS path, not necessarily the root entity type — preventing future readers from incorrectly "fixing" it.

**10. Test `DbContext` instances use `using var` instead of `await using`**

`DynamoComplexPropertyDiscoveryConventionTests.cs` — lines ~55–57, ~71–72, ~87–88

Several tests create `DbContext` with `using var ctx = ...` rather than `await using`. Since `DbContext` is `IAsyncDisposable`, synchronous `using` calls `Dispose` rather than `DisposeAsync`. Typically harmless but inconsistent with the rest of the test suite where `await using` is the project-standard idiom.

---

*Review generated via Claude Code sr-net-reviewer agent.*

ncipollina (changes requested):
Left a few comments.

docs/modeling/complex-types.md
```diff
@@ -0,0 +1,215 @@
...
+  "contacts": {
+    "L": [
+      { "M": { "email": { "S": "ada@example.com" } } }
+    ]
+  }
+}
+```
+
+Supported CLR collection shapes: `List<T>`, `IList<T>`. `ICollection<T>`,
+`IReadOnlyList<T>`, and arrays are not supported for complex collections.
```
ncipollina:
```suggestion
Supported CLR collection shapes: `List<T>`, `IList<T>`. `ICollection<T>`,
and `IReadOnlyList<T>`. Arrays are not supported for complex collections.
```

✏️ This line reads kind of weird

.idea/.idea.EntityFrameworkCore.DynamoDb/.idea/codeStyles/codeStyleConfig.xml
ncipollina, line 1: ❓ Why did you add the `.idea` folder. Please put this in `.gitignore`?

ncipollina (changes requested):
⚠️ Please remove the `.idea` folder.

ncipollina (approved):
🔥 Nice work!!!!

PR #194, opened 2026-05-04
fix(storage): distinguish missing items from concurrency conflicts

### LINKED ISSUE #170: bug: ConditionalCheckFailedException incorrectly maps to DbUpdateConcurrencyException when item does not exist
## Summary

When DynamoDB throws `ConditionalCheckFailedException` during a PartiQL `UPDATE` or `DELETE`, the provider unconditionally maps it to `DbUpdateConcurrencyException`. This causes false positives because DynamoDB raises the same exception whether a concurrency token predicate failed **or** the item simply does not exist.

## Affected code

`src/EntityFrameworkCore.DynamoDb/Storage/DynamoWriteExceptionMapper.cs`

```csharp
if (ex is ConditionalCheckFailedException || IsConditionalCheckFailedException(ex))
    return new DbUpdateConcurrencyException(...); // always, regardless of why
```

## Concrete false positive paths

**1. UPDATE with no concurrency tokens**

The WHERE clause is only `pk = ? [AND sk = ?]`. If the item was externally deleted, DynamoDB throws `ConditionalCheckFailedException` but no concurrency token was ever involved. Mapping this to `DbUpdateConcurrencyException` is incorrect.

**2. UPDATE/DELETE with concurrency tokens where the item was deleted**

The WHERE clause includes `token_col = ?` but the item no longer exists. The actual failure reason is "item not found", not "stale token". EF Core's SQL behaviour treats both as a concurrency conflict (0 rows affected → same outcome), but DynamoDB can distinguish the two cases and produces a misleading exception.

The same issue applies to `ConditionalCheckFailed` cancellation reasons inside `TransactionCanceledException`.

### Option A — Inspect entity type at mapping time (no AWS changes)

In `WrapWriteException`, check whether the entity type has any concurrency token properties:
- No tokens → `DbUpdateException` (item did not exist, no concurrency token was tested)
- Has tokens → `DbUpdateConcurrencyException` (keep current behaviour)

**Pro:** Zero write-path changes. Fixes the clearest false positive (no-token UPDATE/DELETE).  
**Con:** Cannot distinguish "item deleted by another writer" vs. "token changed" when tokens are configured.

### Option B — `ReturnValuesOnConditionCheckFailure = ALL_OLD` + inspect `ex.Item` ✅ preferred

Set `ReturnValuesOnConditionCheckFailure = AllOldAttributes` on every write request (`ExecuteStatementRequest`, `ParameterizedStatement`, `BatchStatementRequest`). On `ConditionalCheckFailedException`:

- `ex.Item` is null/empty → item did not exist → `DbUpdateException` ("item not found; it may have been deleted")
- `ex.Item` has values → item existed but token mismatched → `DbUpdateConcurrencyException`

Same logic for `TransactionCanceledException` via `CancellationReason.Item` per statement.

**Pro:** Precise discrimination with no extra roundtrips. Zero overhead on the success path; only the failure path carries additional data (old item attributes, discarded after the null-check).  
**Con:** Requires threading `ReturnValuesOnConditionCheckFailure` through `ExecuteWriteAsync`, `ExecuteTransactionAsync`, and `ExecuteBatchWriteAsync`.

### Option C — Post-failure `GetItem` probe

On `ConditionalCheckFailedException`, issue a `GetItemAsync` on the primary key:
- Item missing → `DbUpdateException`
- Item present → `DbUpdateConcurrencyException`

**Pro:** Minimal write-path changes.  
**Con:** Extra roundtrip on an already-failed write; small race window between failure and probe.

## Chosen approach

**Option B**, with Option A as a complementary safeguard for the no-token case. Option B uses a DynamoDB-native mechanism designed exactly for this purpose, adds no latency on the happy path, and gives precise exception semantics on failure.

j-d-ha:
## Summary

This updates DynamoDB write failure handling so conditional failures can distinguish missing key targets from actual optimistic concurrency conflicts. The provider now requests DynamoDB `ALL_OLD` data on condition-check failures and uses returned old item presence to classify exceptions more accurately.

## Changes

- Set `ReturnValuesOnConditionCheckFailure = ALL_OLD` for single, transactional, and batch PartiQL writes.
- Map conditional failures with returned old item data to `DbUpdateConcurrencyException`.
- Map conditional failures without returned old item data to `DbUpdateException` with a key-not-found message.
- Preserve batch `BatchStatementError.Item` data when converting batch statement failures.
- Add integration coverage for missing item, wrong sort key, and stale version cases.
- Update concurrency docs to explain key-miss vs token-conflict behavior.

## Validation

- `dotnet build`
- `dotnet test --project tests/EntityFrameworkCore.DynamoDb.Tests/EntityFrameworkCore.DynamoDb.Tests.csproj --no-build`
- `dotnet test --project tests/EntityFrameworkCore.DynamoDb.IntegrationTests/EntityFrameworkCore.DynamoDb.IntegrationTests.csproj --no-build -- --filter-class EntityFrameworkCore.DynamoDb.IntegrationTests.SaveChangesTable.SaveChangesConcurrencyTests`
- `dotnet test --project tests/EntityFrameworkCore.DynamoDb.IntegrationTests/EntityFrameworkCore.DynamoDb.IntegrationTests.csproj --no-build -- --filter-method '*DeleteAsync_NonExistentKey_IsNoOp'`
- `task docs:build`

## Related Issues

Fixes #170

### DIFF
--- a/docs/saving/concurrency.md
+++ b/docs/saving/concurrency.md
@@ -81,8 +81,16 @@ Conflict detection depends on the execution path:
 - **Transactional saves** (`ExecuteTransaction`): DynamoDB returns `TransactionCanceledException`
     with a `ConditionalCheckFailed` cancellation reason for the conflicting item.
 
-The provider maps both to `DbUpdateConcurrencyException`. The `Entries` property on the
-exception identifies which entities were in conflict.
+The provider requests DynamoDB's `ALL_OLD` value on condition-check failures and uses the
+returned item to distinguish key misses from token conflicts:
+
+- If DynamoDB returns the existing item, the key matched and the concurrency token predicate
+    failed; the provider throws `DbUpdateConcurrencyException`.
+- If DynamoDB does not return an item, the target key did not match any item. The item may have
+    been deleted, or the tracked key values may not match the stored item; the provider throws
+    `DbUpdateException`.
+
+The `Entries` property on the exception identifies which entities were involved.
 
 ## Handling DbUpdateConcurrencyException
 
@@ -153,9 +161,10 @@ the same.
 If concurrency tokens are configured, the token value is included in the WHERE predicate. This
 creates an important asymmetry:
 
-- **Item missing entirely** → silent success (provider accepts the delete as done).
-- **Item present with a different token value** → `ConditionalCheckFailedException` →
-    `DbUpdateConcurrencyException`.
+- **Item missing entirely** → silent success for singleton deletes (provider accepts the delete
+    as done) or `DbUpdateException` if DynamoDB reports a failed condition on that write path.
+- **Item present with a different token value** → `ConditionalCheckFailedException` with old item
+    attributes → `DbUpdateConcurrencyException`.
 
 The second case matters when two writers race to delete the same item: the first delete
 succeeds and the item is gone; the second delete finds the item missing and also succeeds. But
--- a/src/EntityFrameworkCore.DynamoDb/Storage/DynamoClientWrapper.cs
+++ b/src/EntityFrameworkCore.DynamoDb/Storage/DynamoClientWrapper.cs
@@ -68,6 +68,8 @@ public class DynamoClientWrapper : IDynamoClientWrapper
                 {
                     Statement = state.statement,
                     Parameters = state.parameters?.Count > 0 ? state.parameters : null,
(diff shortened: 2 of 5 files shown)


ncipollina (approved):
Nice work!!!

PR #200, opened 2026-05-07
feat(query): support FindAsync by primary key

### LINKED ISSUE #175: feat: support FindAsync for single-item retrieval by primary key
## Summary

EF Core's \`FindAsync\` is the idiomatic way to look up a single entity by primary key. It checks the change tracker first and falls back to the store — making it the most efficient single-item retrieval path. Today the provider has no \`IEntityFinder\` override, so \`FindAsync\` falls through to the base EF Core implementation which builds a LINQ predicate and runs it through the full PartiQL pipeline. While this works, EF Core does not guarantee the generated query shape hits the \`First*\` safe path, so it may not benefit from the implicit \`Limit=1\` and single-page optimizations the provider applies to known \`First*\` terminals.

This story wires \`FindAsync\` to an explicit \`IEntityFinder\` implementation that constructs a key-equality predicate and routes it through the existing \`ExecuteStatement\` pipeline under the \`First*\` safe path, ensuring the correct single-page, \`Limit=1\` execution.

> **Note:** Only \`FindAsync\` is in scope. The synchronous \`Find\` is not supported — the provider is async-only.

## Behavior

**PK-only table:**
\`\`\`csharp
var order = await context.Orders.FindAsync(partitionKeyValue);
\`\`\`

**Composite PK + SK table:**
\`\`\`csharp
// Key values must be passed in primary key property order (PK first, SK second)
var line = await context.OrderLines.FindAsync(partitionKeyValue, sortKeyValue);
\`\`\`

Both cases:
1. Check the EF Core change tracker first — tracked entity returned immediately, no network call
2. Build a key-equality LINQ predicate and execute it through the existing PartiQL pipeline via \`ExecuteStatement\` with \`Limit=1\`, single-page execution
3. Return \`null\` if the item does not exist

## Key schema context

The provider already tracks PK and SK metadata via \`DynamoAnnotationNames.PartitionKeyPropertyName\` / \`SortKeyPropertyName\`, surfaced through \`DynamoEntityTypeExtensions.GetPartitionKeyProperty()\` / \`GetSortKeyProperty()\`. The key ordering in EF Core's primary key is rebuilt by \`DynamoKeyInPrimaryKeyConvention\` so it matches DynamoDB's key schema — the \`keyValues\` array passed to \`FindAsync\` arrives in the correct order.

## Implementation sketch

**1. \`DynamoEntityFinder<TEntity>\` (new)**

Implement \`IEntityFinder<TEntity>\`. On a change tracker miss:
- Resolve \`IReadOnlyProperty\` for the partition key and (if present) sort key via \`DynamoEntityTypeExtensions\`
- Build a key-equality \`Expression<Func<TEntity, bool>>\` predicate against the resolved key properties
- Execute via \`AsAsyncEnumerable().FirstOrDefaultAsync()\` so the query routes through the \`First*\` safe path with \`Limit=1\` and single-page execution

**2. DI registration**

Register \`DynamoEntityFinder<>\` as the \`IEntityFinder<>\` open-generic service so EF Core resolves it instead of the default \`EntityFinder<>\`.

## What is NOT in scope

- Synchronous \`Find\` — not supported, provider is async-only
- Routing arbitrary LINQ key-predicate queries (\`Where(x => x.Id == id).FirstOrDefaultAsync()\`) through a dedicated finder — those already work via the existing \`First*\` safe path
- Consistent-read configuration — follow-up story
- Projection support for \`FindAsync\` — not part of the EF Core API

## Acceptance criteria

- [ ] \`FindAsync(pkValue)\` on a PK-only table issues an \`ExecuteStatement\` call with \`Limit=1\`
- [ ] \`FindAsync(pkValue, skValue)\` on a composite-key table issues an \`ExecuteStatement\` call with \`Limit=1\`
- [ ] Returns \`null\` for a missing key (no exception)
- [ ] Returns the tracked instance if already in the change tracker (no network call)
- [ ] Integration tests for both table shapes

j-d-ha:
## Summary

Adds DynamoDB provider support for EF Core `FindAsync` when callers provide the full table primary key. The implementation now matches other EF Core providers by using EF Core's default entity finder: tracked entities are resolved by EF Core first, and untracked lookups flow through the normal async query pipeline as primary-key equality queries with `ExecuteStatementRequest.Limit = 1`.

## Changes

- Uses EF Core's default `IEntityFinderSource` instead of registering a DynamoDB-specific finder wrapper.
- Keeps full base-table primary-key lookups on the base table by short-circuiting automatic secondary-index selection when the predicate fully constrains the table key.
- Preserves ordinary LINQ automatic secondary-index selection for non-`FindAsync` query shapes.
- Adds/updates unit coverage for default finder registration, tracked lookup, `Limit = 1`, missing items, invalid key values, cancellation tokens, and sync `Find` behavior.
- Adds integration coverage proving `FindAsync` uses the base table on indexed tables while ordinary LINQ can still auto-select an index.
- Updates docs for `FindAsync`, base-table lookup behavior, and the sync `Find` tracked-entity exception to async-only DynamoDB I/O.

## Validation

- `dotnet build --no-restore`
- .NET test MCP: `tests/EntityFrameworkCore.DynamoDb.Tests/EntityFrameworkCore.DynamoDb.Tests.csproj` — 461 passed.
- .NET test MCP: `EntityFrameworkCore.DynamoDb.IntegrationTests.SimpleTable.FindAsyncTests` — 2 passed.
- .NET test MCP: `EntityFrameworkCore.DynamoDb.IntegrationTests.PkSkTable.FindAsyncTests` — 2 passed.
- .NET test MCP: `EntityFrameworkCore.DynamoDb.IntegrationTests.SecondaryIndexTable.FindAsyncTests` — 2 passed.
- `uv run zensical build`

## Release Notes

`FindAsync` is now supported for full DynamoDB primary-key lookup. Synchronous `Find` can return an already-tracked entity without DynamoDB I/O, but throws if it needs to execute a synchronous DynamoDB query.

Closes #175


ncipollina:
@codex review

ncipollina (dismissed):
PR looks good, but it is out of sync w/ main.

ncipollina (approved):
🚢 ship it!

PR #203, opened 2026-05-11
feat(metadata): add key resolution and transaction handling

j-d-ha:
## Summary

Moves the non-test functional work from the specification-test branch into a focused PR against `main`. This adds centralized key resolution/table-group handling and wires unsupported transaction handling so functional fixes can land before continuing integration/specification test work.

## Changes

- Add metadata helpers for resolving key properties and computing table group names.
- Streamline table-group usage across query translation, PartiQL statement creation, save planning, and transaction target identity creation.
- Add `DynamoTransactionManager` to explicitly handle unsupported transaction APIs.
- Fix solution folder path from `/git/` to `/.github/`.
- Update agent guidance for NuGet package management through the `dotnet` CLI.

## Validation

- `dotnet build`
- `dotnet test --project tests/EntityFrameworkCore.DynamoDb.Tests/EntityFrameworkCore.DynamoDb.Tests.csproj --no-build`

## Notes for Reviewers

Specification/integration test work remains on `feat/tests-specification-tests-v2` for follow-up after these functional changes land.

### DIFF
--- a/src/EntityFrameworkCore.DynamoDb/Storage/DynamoDatabaseCreator.cs
+++ b/src/EntityFrameworkCore.DynamoDb/Storage/DynamoDatabaseCreator.cs
@@ -1,0 +1,64 @@
+using Microsoft.EntityFrameworkCore.Storage;
+
+namespace EntityFrameworkCore.DynamoDb.Storage;
+
+/// <summary>Provides unsupported database lifecycle operations for the DynamoDB provider.</summary>
+internal sealed class DynamoDatabaseCreator : IDatabaseCreator
+{
+    private const string DatabaseLifecycleNotSupported =
+        "The DynamoDB database provider does not support database lifecycle operations.";
+
+    /// <summary>Ensures the database is deleted.</summary>
+    /// <returns>Never returns because database lifecycle operations are unsupported.</returns>
+    /// <exception cref="NotSupportedException">
+    ///     Always thrown because database lifecycle operations are
+    ///     unsupported.
+    /// </exception>
+    public bool EnsureDeleted() => throw new NotSupportedException(DatabaseLifecycleNotSupported);
+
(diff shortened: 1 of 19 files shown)


src/EntityFrameworkCore.DynamoDb/Storage/DynamoDatabaseCreator.cs
ncipollina, line 1: ❓ Assuming we will fill this out in a future PR?

src/EntityFrameworkCore.DynamoDb/Storage/DynamoDatabaseWrapper.cs
```diff
@@ -53,6 +57,9 @@ public override async Task<int> SaveChangesAsync(
 
         try
         {
+            if (Transaction.Current is not null)
```
ncipollina, line 60: ❓ We aren't supporting transactions?

ncipollina (approved):
In general ok with this change. Did leave a few questions, but I'm assuming you are planning on addressing in future PRs

src/EntityFrameworkCore.DynamoDb/Storage/DynamoDatabaseWrapper.cs
j-d-ha: We do, but only for writes. Dynamo transactions are just an array of changes to be made at once and it will then roll back if it fails. It does not support more complicated transaction semantics like a relational DB.

src/EntityFrameworkCore.DynamoDb/Storage/DynamoDatabaseCreator.cs
j-d-ha: Yup, just needed this as a place holder for a test.

PR #213, opened 2026-05-13
fix(query): preserve nullable value projection nulls

j-d-ha:
## Summary

Fixes nullable value-type projection materialization so member chains rooted at `Nullable<T>.Value` preserve database nulls instead of evaluating `.Value` on a null nullable. This keeps projected nullable results, such as `DateTimeOffset?.Value.DateTime.Date` or `int?.Value`, null-safe after DynamoDB projection binding removal.

## Changes

- Added a projection-binding removal rewrite for nullable value member/method chains converted to nullable result types.
- The rewrite materializes the nullable source once, checks it for null, and only evaluates the `.Value` chain on non-null values.
- Added SimpleTable integration coverage for nullable value projection null preservation.

## Validation

- `dotnet test --project tests/EntityFrameworkCore.DynamoDb.IntegrationTests/EntityFrameworkCore.DynamoDb.IntegrationTests.csproj -- --filter-method '*Select_NullableValueTypeValueProjection_PreservesNull'`

## Notes for Reviewers

This branch is based on `main`. The earlier specification-test unskip from the working branch was not included because those specification test source files are not present on `main`.


ncipollina (approved):
✅ LGTM.

PR #215, opened 2026-05-13
fix(metadata): configure Dynamo value generation convention

j-d-ha:
## Summary

Adds a DynamoDB-specific value-generation convention so EF Core no longer marks numeric or string keys as generated by convention. This matches non-relational provider behavior: DynamoDB does not have identity columns, while Guid keys can continue using EF Core client-side generation.

## Changes

- Register `DynamoValueGenerationConvention` in the provider convention set.
- Keep `Guid` key generation via EF Core client-side defaults.
- Leave non-Guid keys application-assigned unless users explicitly configure generation.
- Add dedicated convention tests for `int` and `Guid` partition keys.

## Validation

- `dotnet test --project tests/EntityFrameworkCore.DynamoDb.Tests/EntityFrameworkCore.DynamoDb.Tests.csproj -- --filter-class '*DynamoKeyInPrimaryKeyConventionTests'`\n- `dotnet test --project tests/EntityFrameworkCore.DynamoDb.Tests/EntityFrameworkCore.DynamoDb.Tests.csproj -- --filter-class '*DynamoValueGenerationConventionTests'`\n- `dotnet build --no-restore`\n\n## Notes for Reviewers\n\nThis mirrors Cosmos/Mongo-style non-relational provider conventions: suppress default integer key value generation and only preserve supported client-side generation patterns.

### DIFF
--- a/docs/modeling/entities-keys.md
+++ b/docs/modeling/entities-keys.md
@@ -69,8 +69,12 @@ If a numeric key is left unset, EF Core treats the CLR default (for example `0`)
 write. If a string key is left `null`, the save fails because DynamoDB key attributes must be
 present and non-null.
 
-`Guid` keys keep EF Core's default client-side generation behavior. Leaving a `Guid` key as
-`Guid.Empty` lets EF Core assign a new Guid before writing the item.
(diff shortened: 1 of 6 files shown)


ncipollina (approved):
👍 Good to go!!

PR #216, opened 2026-05-14
feat(metadata): add Id fallback for partition key convention

j-d-ha:
## Summary

Adds `Id` / `ID` as a **fallback** partition-key convention for root DynamoDB entities. The provider still prefers DynamoDB-specific names (`PK` / `PartitionKey`) and still treats `HasPartitionKey(...)` as the authoritative explicit API.

This is intended to reduce friction for simple models and tests where `Id` is already the natural key name, without opening the larger `HasKey(...)` / `[Key]` inference path. This is not a breaking change for released users because the provider has not shipped this behavior yet, but it is a meaningful convention change and should be reviewed carefully.

## Rationale

Before this change, convention-based table-key discovery only recognized `PK` / `PartitionKey` for partition keys and `SK` / `SortKey` for sort keys. That matches common DynamoDB single-table modeling, but it makes simple EF-style entities noisy because many models naturally start with an `Id` property and then require explicit `HasPartitionKey(x => x.Id)` configuration.

Other EF providers commonly support `Id` as a key convention:

- EF Core base conventions discover `Id` / `<EntityName>Id` as primary keys.
- Cosmos builds on EF key discovery while separately handling partition keys and JSON `id` mapping.
- MongoDB provider falls back to EF-style `Id` discovery and maps the chosen key to `_id`.

DynamoDB table keys are provider-specific, so this PR deliberately does **not** adopt the full EF `HasKey(...)` / `[Key]` inference model. Instead it adds the narrowest useful convention: `Id` only works when no more-specific DynamoDB key name exists.

## Behavior

Partition-key discovery now resolves in this order:

1. Explicit `HasPartitionKey(...)`
2. Conventional DynamoDB names: `PK` / `PartitionKey`
3. Fallback conventional name: `Id` / `ID`

Sort-key discovery is unchanged:

- `SK` / `SortKey`
- Explicit `HasSortKey(...)` still wins

`HasKey(...)` and `[Key]` are still rejected for root DynamoDB entities. The provider continues to derive the EF primary key from the resolved DynamoDB partition/sort key schema.

## Why fallback, not equal alternative

`Id` is intentionally a fallback instead of an equal peer to `PK` / `PartitionKey`.

If `Id` were treated as an equal partition-key convention, an entity with both `PK` and `Id` would become ambiguous. That is too broad because many DynamoDB models use:

- `PK` as the actual DynamoDB partition key
- `Id` as a separate domain identifier

With fallback semantics:

```csharp
public class Item
{
    public string PK { get; set; } = null!;
    public string Id { get; set; } = null!;
}
```

resolves to:

- partition key: `PK`
- `Id`: normal property

while:

```csharp
public class Item
{
    public string Id { get; set; } = null!;
}
```

resolves to:

- partition key: `Id`
- EF primary key: `[Id]`

and:

```csharp
public class Item
{
    public string Id { get; set; } = null!;
    public string SK { get; set; } = null!;
}
```

resolves to:

- partition key: `Id`
- sort key: `SK`
- EF primary key: `[Id, SK]`

`PK` + `PartitionKey` remains ambiguous and still requires `HasPartitionKey(...)` to resolve.

## Implementation Notes

- Added shared candidate selection in `DynamoKeyDiscoveryConvention.GetPartitionKeyCandidates(...)`.
- Both key discovery and annotation finalization use the same priority rule.
- Candidate selection walks mapped properties during model building only; there is no query/runtime cost.
- The implementation avoids `<EntityName>Id` because that would bring in a much broader EF convention surface than intended.

## Changes

- Added fallback `Id` / `ID` partition-key discovery.
- Preserved `PK` / `PartitionKey` precedence over `Id`.
- Preserved explicit `HasPartitionKey(...)` precedence over all conventions.
- Preserved rejection of `HasKey(...)` and `[Key]` for root entities.
- Updated validation messages to mention fallback `Id`.
- Added tests for:
  - `Id` fallback
  - `ID` fallback
  - `Id + SK` composite key
  - explicit `HasPartitionKey(...)` overriding `Id`
  - `PK` beating fallback `Id`
- Updated docs in:
  - `docs/configuration/table-key-mapping.md`
  - `docs/modeling/entities-keys.md`

## Validation

- `dotnet test --project tests/EntityFrameworkCore.DynamoDb.Tests/EntityFrameworkCore.DynamoDb.Tests.csproj --no-restore`
- `dotnet build --no-restore`
- `task docs:build`

## Release Notes

Root entities can now use an `Id` / `ID` property as the DynamoDB partition key by convention when no `PK` or `PartitionKey` property is present. `HasPartitionKey(...)` remains the explicit override, and `HasKey(...)` / `[Key]` remain unsupported for root DynamoDB table-key mapping.

## Notes for Reviewers

Please focus review on whether the fallback semantics are the right line:

- Should `Id` be fallback only, as implemented?
- Should we continue avoiding `<EntityName>Id`?
- Are the docs clear enough that `Id` maps to the physical DynamoDB attribute name `Id` unless users configure `HasAttributeName(...)`?


ncipollina (commented):
✅ LGTM

PR #219, opened 2026-05-14
fix(metadata): allow nullable attribute name mappings

### LINKED ISSUE #217: [Bug]: HasAttributeName warns for nullable reference properties
## What happened?

Using `HasAttributeName(...)` on an optional nullable reference property produces a C# nullability warning:

```text
Nullability of type argument 'string?' must match 'notnull' constraint in order to use it as parameter 'TProperty'.
```

The mapping is valid because the DynamoDB attribute is optional/nullable, but the extension method's generic constraints require a non-null `TProperty`, so mapping a nullable string via the strongly typed property expression warns.

Expected: optional nullable reference properties should be mappable with `HasAttributeName(...)` without requiring consumers to use null-forgiving operators, string-based property configuration, or warning suppressions.

A workaround is to use string-based property configuration:

```csharp
builder.Property<string>(nameof(LeaderboardEntry.ImageUrl)).HasAttributeName("imageUrl");
```

However, the more natural EF mapping emits the warning:

```csharp
builder.Property(entry => entry.ImageUrl).HasAttributeName("imageUrl");
```

## Steps to reproduce

1. Create an entity with a nullable reference property:

```csharp
public sealed class LeaderboardEntry
{
    public string Pk { get; set; } = null!;
    public string Sk { get; set; } = null!;
    public string? ImageUrl { get; set; }
}
```

2. Configure the entity with a typed nullable property expression and `HasAttributeName(...)`:

```csharp
modelBuilder.Entity<LeaderboardEntry>(builder =>
{
    builder.ToTable("table-name");
    builder.HasPartitionKey(entry => entry.Pk);
    builder.HasSortKey(entry => entry.Sk);
    builder.Property(entry => entry.ImageUrl).HasAttributeName("imageUrl");
});
```

3. Build with nullable reference types enabled.

4. Observe the compiler warning:

```text
Nullability of type argument 'string?' must match 'notnull' constraint in order to use it as parameter 'TProperty'.
```

## Environment

- Provider: `EntityFrameworkCore.DynamoDb` `1.0.0-alpha.37`
- .NET: `net10.0`
- Nullable reference types: enabled
- OS: macOS

ncipollina:
## 📋 Summary

Fixes nullable reference property mappings for `HasAttributeName` so calls like `builder.Property(x => x.ImageUrl).HasAttributeName("imageUrl")` no longer produce CS8714. The generic fluent extension now avoids leaking EF Core's nullable `notnull` generic constraint to consumers while preserving the existing type-preserving return shape.

## 📝 Changes

- Suppressed nullable analysis only around the generic `PropertyBuilder<TProperty>.HasAttributeName(...)` extension declaration.
- Added a regression test covering `HasAttributeName` on a nullable `string?` property and verifying the mapped DynamoDB attribute name is applied.

## 🧪 Validation

- `dotnet build EntityFrameworkCore.DynamoDb.slnx -c Release --no-restore`
- `dotnet test --project tests/EntityFrameworkCore.DynamoDb.Tests/EntityFrameworkCore.DynamoDb.Tests.csproj -c Release --no-build --no-restore`

## 🧩 Related Issues

Closes #217

## 📦 Release Notes

Nullable reference properties can now use the generic `HasAttributeName` fluent API without CS8714 nullable constraint warnings.

### DIFF
--- a/src/EntityFrameworkCore.DynamoDb/Extensions/DynamoPropertyBuilderExtensions.cs
+++ b/src/EntityFrameworkCore.DynamoDb/Extensions/DynamoPropertyBuilderExtensions.cs
@@ -21,6 +21,9 @@ public static class DynamoPropertyBuilderExtensions
         }
     }
 
+    // #nullable disable removes the implicit notnull constraint on TProperty so callers
+    // mapping nullable reference properties (e.g. string?) do not get CS8714.
+#nullable disable
     extension<TProperty>(PropertyBuilder<TProperty> propertyBuilder)
     {
         /// <summary>Configures the DynamoDB attribute name used to store this scalar property.</summary>
@@ -29,6 +32,7 @@ public static class DynamoPropertyBuilderExtensions
             => (PropertyBuilder<TProperty>)((PropertyBuilder)propertyBuilder)
                 .HasAttributeName(name);
     }
+#nullable restore
 
     extension(ComplexCollectionTypePropertyBuilder builder)
     {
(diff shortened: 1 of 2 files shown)


j-d-ha (approved):
Love this!

PR #224, opened 2026-05-15
fix(query): parameterize binary constants

j-d-ha:
## Summary

Inline PartiQL literal generation does not support binary values, so queries comparing a binary property to an inline `byte[]` constant could fail during SQL generation. This change emits a PartiQL parameter for binary constants instead, preserving binary query support without requiring callers to hoist values into variables.

## Changes

- Parameterize `byte[]` SQL constants in `DynamoQuerySqlGenerator` instead of attempting inline literal generation.
- Add a focused `ToQueryString` unit test that verifies inline binary constants produce `?` plus a binary parameter comment.

## Validation

- `dotnet test --project tests/EntityFrameworkCore.DynamoDb.Tests/EntityFrameworkCore.DynamoDb.Tests.csproj -- --filter-class EntityFrameworkCore.DynamoDb.Tests.Query.ToQueryStringTests`


ncipollina (approved):
✅ LGTM

PR #227, opened 2026-05-16
feat(dynamodb): support complex type tracking scenarios

j-d-ha:
## Summary

Adds DynamoDB support and coverage for EF Core complex type tracking scenarios while also hardening related serialization and query behavior. This PR includes more than the new specification test: it updates write-plan serialization for complex/nested values, adds unsafe filtered query escape hatches, and documents new behavior and limitations.

### Complex type tracking coverage

- Added `ComplexTypesTrackingDynamoTest` specification coverage for complex property and collection state changes.
- Added DynamoDB-specific skip reasons for unsupported EF specification scenarios.
- Filled in transaction, nullability, default-value, record collection, and array collection assertions.

### DynamoDB write serialization

- Reworked `DynamoEntityWritePlanFactory` scalar and collection serialization so complex scalar properties and property-converted collections are written correctly.
- Added tests for complex scalar values, converted collection properties, and nested item serialization behavior.
- Documented complex/nested serialization details in saving docs.

### Query safety and pagination behavior

- Added `AllowUnsafeFilteredFirstWithoutOrderBy` configuration on `DynamoDbContextOptionsBuilder` plus queryable extension support.
- Threaded unsafe filtered `First*` handling through query compilation/postprocessing.
- Expanded tests around `FirstOrDefault`, safe-path validation, filtered queries, and pagination configuration.

### Projection cleanup

- Simplified projection binding removal assignment logic in query visitor code.

### Documentation

- Updated DbContext configuration docs for unsafe filtered query options.
- Updated querying operator, ordering/limiting, limitations, and saving docs so user-facing guidance matches provider behavior.

## Validation

- `dotnet test --project tests/EntityFrameworkCore.DynamoDb.Tests/EntityFrameworkCore.DynamoDb.Tests.csproj --no-restore` — passed: 607 succeeded, 2 skipped.

## Notes for Reviewers

Review by bucket rather than only by the new specification test: complex write serialization, unsafe filtered query behavior, docs, and spec-test enablement each have separate behavior implications.


ncipollina:
@codex review

ncipollina (commented):
## Code Review — `test/add-ComplexTypesTrackingTest`

Overall the PR is solid: the new `AsUnsafeFilteredQuery` / `AllowUnsafeFilteredQueries` surface is well-layered, the unit tests cover the interesting edge cases, and the `ComplexTypesTrackingDynamoTest` spec harness is a good first pass. The serialization expansion in `DynamoEntityWritePlanFactory` is the most complex change and has a few correctness and maintainability issues worth addressing.

---

### 1. `DynamoEntityWritePlanFactory` — empty-set sentinel may hide data silently

**File:** `src/EntityFrameworkCore.DynamoDb/Storage/DynamoEntityWritePlanFactory.cs`

In `SerializeBoxedScalarSet`, when the input enumerable is empty (zero elements), the method returns `new AttributeValue { NULL = true }`. This conflates "the set has no items" with "the property value is null". DynamoDB does not allow empty sets, so the correct behavior here is to throw rather than silently write `NULL`, or to document the deliberate downgrade clearly with a caller-facing exception. As it stands a consumer can write an empty `HashSet<string>` and read back `null`, which is a silent data-loss/type-confusion bug.

```csharp
// Current (line ~382):
return ns is { Count: > 0 }
    ? new AttributeValue { NS = ns }
    : new AttributeValue { NULL = true };   // ← empty set silently becomes NULL
```

**Suggested fix:** throw `InvalidOperationException("DynamoDB sets cannot be empty; use a null property or a non-empty collection.")` when all three lists remain null/empty after the loop.

---

### 2. `DynamoEntityWritePlanFactory` — `MemoryStream` wrapping `byte[]` set elements is never disposed

**File:** `src/EntityFrameworkCore.DynamoDb/Storage/DynamoEntityWritePlanFactory.cs`

`SerializeBoxedScalarSet` allocates `new MemoryStream(bytes, false)` for each binary element and adds it to `bs`. These `MemoryStream` instances are handed to `AttributeValue { BS = bs }` and are never disposed. While `MemoryStream` over an existing buffer does not hold unmanaged resources and GC will collect it, this is a code-smell that will trigger disposability analyzers and is inconsistent with the rest of the codebase's resource hygiene.

---

### 3. `DynamoEntityWritePlanFactory` — `SerializeDirectScalarDictionary` type-dispatch list is incomplete vs `SerializeDirectScalarList`

**File:** `src/EntityFrameworkCore.DynamoDb/Storage/DynamoEntityWritePlanFactory.cs`

`SerializeDirectScalarList` handles 16 numeric CLR types (bool, byte, sbyte, short, ushort, int, uint, long, ulong, float, double, decimal + nullable variants + byte[]). `SerializeDirectScalarDictionary` handles only 7 value types (int, int?, long, long?, decimal, decimal?, string) before falling back to `SerializeBoxedScalarDictionary`. This asymmetry means a `Dictionary<string, bool>` inside a complex property will hit the boxed slow path, which is fine for correctness but is inconsistent and may produce surprising behavior for numeric types where the boxed `ConvertProviderValueToAttributeValue` path might not produce a number string. The more pressing correctness risk is that `ReadKeyValuePair`, which is the boxed dictionary path, also only handles 7 value types and throws `NotSupportedException` for everything else — so `Dictionary<string, bool>` would actually throw at runtime, not just be slow.

The fix is either to expand both fast-path dispatch tables to cover the same type set as `SerializeDirectScalarList`, or document the supported dictionary value types and validate them at model-building time.

---

### 4. `DynamoQueryTranslationPostprocessor` — `HasUserLimit` check moved outside `ValidateFirstTerminalSafePath`, bypasses `unsafeFilteredQueriesAllowed`

**File:** `src/EntityFrameworkCore.DynamoDb/Query/Internal/DynamoQueryTranslationPostprocessor.cs`

The refactor correctly separates the `HasUserLimit` guard from the safe-path validator, but placing it *before* the `unsafeFilteredQueriesAllowed` check means `Limit(n) + First*` always throws — even when `AsUnsafeFilteredQuery()` is active. This is the intended behavior per the docs ("does not change `Limit(n)` semantics"), and the corresponding unit test `FirstOrDefault_NonKeyFilter_WithUnsafeFilteredQueryAndUserLimit_ThrowsTranslationFailure` confirms it. That is good. However, the comment block that used to accompany this guard in `ValidateFirstTerminalSafePath` was removed entirely (diff lines 348-352). This was the only place explaining *why* `Limit(n) + First*` is unconditionally rejected. The rationale ("ambiguous — Limit(n) may return multiple items") should be preserved as a comment on the new early-return location so the intent is not lost during future refactors.

---

### 5. `ComplexTypesTrackingDynamoTest` — sync overrides call `base.*` directly, not `NoSyncTest`

**File:** `tests/EntityFrameworkCore.DynamoDb.SpecificationTests/ComplexTypesTrackingDynamoTest.cs`

The `AGENTS.md` / spec-test CLAUDE.md for the SpecificationTests project states:

> **Sync methods** — DynamoDB has no sync query path:
> ```csharp
> public override void Some_sync_test()
>     => NoSyncTest(() => base.Some_sync_test());
> ```

Many synchronous overrides in this file (e.g. `Can_mark_complex_type_properties_modified`, `Detect_changes_in_complex_type_properties`, `Can_read_original_values_for_properties_of_complex_types`, etc.) simply delegate straight to `base.*()` without wrapping in `NoSyncTest`. If those base tests exercise any DynamoDB query paths they will fail or hang. Tracking-only tests that never issue queries may be safe, but the pattern should be consistent with the rest of the spec test suite.

---

### 6. `ComplexTypesTrackingDynamoTest` — array-backed entities silently ignored rather than skipped with a reason

**File:** `tests/EntityFrameworkCore.DynamoDb.SpecificationTests/ComplexTypesTrackingDynamoTest.cs` (line ~1055-1059)

The fixture ignores `PubWithArrayCollections` and `PubWithRecordArrayCollections` via `modelBuilder.Ignore<...>()`. This causes the corresponding `Can_track_entity_with_complex_type_array_collections`, `Can_track_entity_with_complex_struct_array_collections`, etc. overrides to silently pass (they call `base.*` which will find no entities). A reader has no indication that these tests are structurally limited rather than genuinely passing. The `SkipReason.ArrayComplexCollectionsNotSupported` constant exists for exactly this case — consider applying it with `[ConditionalTheory(Skip = SkipReason.ArrayComplexCollectionsNotSupported)]` on the array-collection overrides instead of ignoring the model types.

---

### 7. `ComplexTypesTrackingDynamoTest.ExecuteWithStrategyInTransactionAsync` — outer context is created but never used for data operations

**File:** `tests/EntityFrameworkCore.DynamoDb.SpecificationTests/ComplexTypesTrackingDynamoTest.cs` (lines ~965-1019)

`ExecuteWithStrategyInTransactionAsync` creates an outer `context` for the purpose of calling `CleanAsync`, then creates a fresh `innerContext` for each test operation. This is the correct isolation pattern. One concern: if `testOperation` throws, `CleanAsync` in the `finally` block still runs but operates on the *outer* context (which has never been used for data). That is correct. However, `CreateExecutionStrategy().ExecuteAsync` wraps only the test operations and leaves `CleanAsync` outside the retry boundary. A transient DynamoDB Local failure during cleanup will propagate directly. This is an accepted trade-off in most EF provider spec tests but worth noting if flakiness is observed.

---

### 8. `DynamoDbQueryableExtensions` — `AsUnsafeFilteredQuery` XML doc line exceeds 120 characters

**File:** `src/EntityFrameworkCore.DynamoDb/Extensions/DynamoDbQueryableExtensions.cs` (lines ~187-188)

The `<remarks>` block has a line that exceeds the project's line-length guide:

```
///     This does not disable scan-like query protection, does not allow <c>WithNextToken(...)</c>
///     with <c>First*</c>, and does not change <c>First*</c> execution semantics: the provider still
```

The second sentence runs long. Minor, but consistent with the project's formatting standards.

---

### 9. `DynamoEntityItemSerializerSourceTests` — `BuildItem_ComplexTypeWithScalarCollections_SerializesListsAndBinary` does not assert `convertedCodes` default

**File:** `tests/EntityFrameworkCore.DynamoDb.Tests/Storage/DynamoEntityItemSerializerSourceTests.cs`

In `BuildItem_ComplexTypeWithScalarCollections_SerializesListsAndBinary` the entity sets `ConvertedCodes = new ConvertedCodeList(string.Empty)` implicitly via the field initializer. The test does not assert the resulting DynamoDB shape for `convertedCodes`. This is intentional separation (the converter test is in the next test method), but a comment noting the deliberate omission would help readers understand why every field is asserted except `convertedCodes`.

---

### 10. `SkipReason.ComplexCollectionScanMaterializationNotSupported` — constant is declared but not referenced

**File:** `tests/EntityFrameworkCore.DynamoDb.SpecificationTests/SkipReason.cs`

`ComplexCollectionScanMaterializationNotSupported` is added but no override in `ComplexTypesTrackingDynamoTest` uses it as a `[ConditionalTheory(Skip = ...)]` value. Either remove it or apply it to the overrides that were supposed to use it (possibly the `Can_change_state_from_Deleted_with_complex_*` family that skip sync via `Task.CompletedTask` without explanation).

ncipollina (changes requested):
I left some feedback. Take a look and respond.

ncipollina (commented):
## Re-review — `test/add-ComplexTypesTrackingTest`

**Previous issues resolved (thank you):**
- `SerializeBoxedScalarSet` now throws on empty set (was: silent NULL) — fixed
- Dictionary serialization now covers all 16 value types via full dispatch — fixed
- `MemoryStream` leak centralized into `CreateBinaryStream` — fixed
- Comment explaining why `Limit(n) + First*` is rejected added to postprocessor — fixed
- `convertedCodes` assertion added to serializer test — fixed
- `SkipReason.ComplexCollectionScanMaterializationNotSupported` removed — fixed

**Three issues remain from the previous pass. Details inline below.**

---

### Issue A (still present) — `modelBuilder.Ignore<>()` silently masks array-backed collection tests

`ComplexTypesTrackingDynamoTest.ComplexTypesTrackingDynamoFixture.OnModelCreating` calls `modelBuilder.Ignore<PubWithArrayCollections>()` and `modelBuilder.Ignore<PubWithRecordArrayCollections>()`. The corresponding test overrides (`Can_track_entity_with_complex_type_array_collections`, `Can_track_entity_with_complex_struct_array_collections`, `Can_track_entity_with_complex_readonly_struct_array_collections`, `Can_track_entity_with_complex_record_array_collections`) all delegate to `base.Can_track_entity_with_complex_*_array_collections(state, async)` when `async == true`. Because the entity types are ignored at the model level, the base test body silently produces zero rows rather than verifying anything useful — it passes by vacuous success rather than by actually exercising the limitation.

**Recommended fix:** Decorate each array-backed `async`/`true` override with `[ConditionalTheory(Skip = SkipReason.ArrayComplexCollectionsNotSupported)]` (restore the `ArrayComplexCollectionsNotSupported` constant that was deleted in commit `4454b5a`, or add a local inline literal if preferred), and return `Task.CompletedTask` unconditionally. That makes the skip intentional and machine-readable rather than accidentally passing.

**Alternatively**, if the base test already guards for an empty context and the intent is "these tests pass harmlessly against an empty model", add a comment explaining that explicitly and keep the current shape. The current in-code comment (`// DynamoDB complex collections currently require List<T>/IList<T>; ignore array-backed`) is in the fixture, not visible at the override callsite.

---

### Issue B (still present) — `ExecuteWithStrategyInTransactionAsync` cleanup outside retry boundary

In `ComplexTypesTrackingDynamoTest.ExecuteWithStrategyInTransactionAsync` (line ~861), the `finally` block runs `CleanAsync` on a context that was created **before** the `ExecuteAsync` call and is therefore outside the execution-strategy retry scope:

```csharp
await using var context = CreateContext();            // outer context, not retried
await Fixture.TestStore.CleanAsync(context);         // pre-clean outside retry

try
{
    await context.Database.CreateExecutionStrategy().ExecuteAsync(context, async _ =>
    {
        // inner contexts per operation — correct
    });
}
finally
{
    await Fixture.TestStore.CleanAsync(context);    // ← post-clean outside retry boundary
}
```

If the execution strategy retries, the `finally` cleanup runs exactly once after the final attempt (correct for post-test teardown), but the pre-clean at line 811 is also outside the retry boundary. More importantly, all inner `DbContext` instances use `AutoTransactionBehavior.Never`, which is fine, but the outer `context` object is left open across the whole call and could have stale state on a retry. This matches the pattern flagged in the previous review and is unchanged.

**Recommended fix:** Move `CleanAsync` inside the retry lambda (run before each attempt) so state is guaranteed clean on every retry. The post-clean `finally` can remain as-is since it is purely for test isolation cleanup after a final failure.

---

### Issue C (still present) — XML doc `<exception cref="...">` lines exceed 120-char limit

Several lines in `src/EntityFrameworkCore.DynamoDb/Extensions/DynamoDbQueryableExtensions.cs` (newly added in this branch) exceed 120 characters, violating the `.editorconfig` `max_line_length = 120` rule for XML:

- Line 30: 127 chars (`<param name="nextToken">An optional continuation token…</param>`)
- Line 34: 128 chars (`<exception cref="ArgumentOutOfRangeException">…</exception>`)
- Line 79: 144 chars (`<exception cref="ArgumentNullException">…source…nextToken…</exception>`)
- Line 80: 125 chars (`<exception cref="ArgumentException">…empty or whitespace…</exception>`)
- Line 246: 144 chars (`<exception cref="ArgumentNullException">…source…indexName…</exception>`)

**Recommended fix:** Wrap each long XML doc tag across two lines, e.g.:
```xml
/// <exception cref="ArgumentNullException">
///     Thrown when <paramref name="source" /> or <paramref name="nextToken" /> is null.
/// </exception>
```

---

**No new issues found.** The serialization logic, async patterns, structured logging, guard clauses, test coverage, and documentation additions all look correct. The `AsUnsafeFilteredQuery` / `AllowUnsafeFilteredQueries` implementation is clean and the `FirstOrDefaultSafePathTests` suite covers the contract thoroughly.

ncipollina (approved):
✅ LGTM!!!

PR #230, opened 2026-05-19
test(tests): add DynamoDB concurrency spec coverage

j-d-ha:
## Summary

Adds DynamoDB specification coverage for optimistic concurrency and concurrency detector behavior. Also narrows auto index-selection diagnostics so non-keyed reads are classified by scan diagnostics instead of misleading no-compatible-index warnings.

## Changes

- Added DynamoDB concurrency spec tests for duplicate add, stale update/delete, derived entities, and relationship cases.
- Added enabled/disabled concurrency detector spec test coverage.
- Centralized DynamoDB test-store cleanup support for concurrency fixtures.
- Updated auto index-selection analyzer to suppress per-index rejection diagnostics when predicates do not target secondary index partition keys.
- Refreshed query diagnostics and index-selection docs to match scan-vs-index warning behavior.

## Validation

- Not run in this session.

## Release Notes

- Query diagnostics now avoid no-compatible-index warnings when a query does not target any secondary index partition key; scan diagnostics remain responsible for non-keyed reads.

### DIFF
--- a/docs/diagnostics.md
+++ b/docs/diagnostics.md
@@ -241,22 +241,24 @@ Index selection runs when [automatic index selection](querying/index-selection.m
 events except `DYNAMO_IDX004` (explicit `.WithIndex(...)` hints) and `DYNAMO_IDX006`
 (explicit `.WithoutIndex()` hints), which always fire.
 
-When multiple candidates are evaluated, `DYNAMO_IDX005` rejection events appear first, followed by
-the final outcome event (`DYNAMO_IDX001`, `DYNAMO_IDX002`, or `DYNAMO_IDX003`). Enable
-`Information` level logging for the `Query` category to see the full decision trail.
+When multiple targeted candidates are evaluated and rejected, `DYNAMO_IDX005` rejection events
+appear first, followed by the final outcome event (`DYNAMO_IDX001`, `DYNAMO_IDX002`, or
+`DYNAMO_IDX003`). Candidates whose partition key is not constrained by the query are not treated as
+rejections; scan-like base-table reads are reported by `ScanLikeQueryDetected` instead.
+Enable `Information` level logging for the `Query` category to see the full decision trail.
 
 ______________________________________________________________________
 
 ### `DYNAMO_IDX001` — `NoCompatibleSecondaryIndexFound` (Warning)
 
-No secondary index on the queried table satisfies the predicate. The query will use the base
-table.
+At least one secondary index was targeted by the predicate, but every targeted candidate failed a
+later safety gate. The query will use the base table.
 
-This is a warning rather than an informational event because a base-table query may scan more data
-than you intended. The message also reminds you to ensure the WHERE clause includes an equality
-constraint on an index partition key if you want automatic selection to pick up a GSI. Review
-whether a GSI would allow the provider to target a narrower key space, and confirm that running
-against the base table is acceptable for this query's access pattern.
+This event is not emitted for unkeyed reads such as `ToListAsync()` or predicates that do not
+constrain any configured secondary-index partition key; those shapes are handled by
+`ScanLikeQueryDetected`. Review the preceding `DYNAMO_IDX005` rejection diagnostics to see why the
+targeted index candidates were rejected, then either adjust the query/index shape or confirm that
+running against the base table is acceptable for this access pattern.
 
 Use EF Core warning configuration to escalate or suppress this warning:
 
--- a/docs/diagnostics.md
+++ b/docs/diagnostics.md
@@ -320,21 +320,21 @@ ______________________________________________________________________
 
 ### `DYNAMO_IDX005` — `SecondaryIndexCandidateRejected` (Information)
 
-A candidate index was evaluated and eliminated. The rejection reason is included in the message:
+A targeted candidate index was evaluated and eliminated. The rejection reason is included in the message:
 
 ```
 info: Microsoft.EntityFrameworkCore.Query[30108]
-      Index 'ByCategory-index' on table 'Orders' was rejected: no equality or IN constraint on the index partition key.
+      Index 'ByCategory-index' on table 'Orders' was rejected: predicate contains an unsafe OR that would produce incorrect results.
 ```
 
 Common rejection reasons:
 
-- No equality or `IN` constraint on the index partition key
 - Predicate contains an unsafe `OR` that would produce incorrect results
 - Index projection type is not `ALL` (only ALL projection is auto-selected)
 
-These events appear before the final outcome event and give you visibility into why specific
-indexes were ruled out.
+Indexes whose partition key is not constrained by equality or `IN` are skipped silently rather than
+reported as rejections. These events appear before the final outcome event and give you visibility
+into why specific indexes were ruled out.
 
 ______________________________________________________________________
 
--- a/docs/diagnostics.md
+++ b/docs/diagnostics.md
@@ -244,7 +244,8 @@ events except `DYNAMO_IDX004` (explicit `.WithIndex(...)` hints) and `DYNAMO_IDX
 When multiple targeted candidates are evaluated and rejected, `DYNAMO_IDX005` rejection events
 appear first, followed by the final outcome event (`DYNAMO_IDX001`, `DYNAMO_IDX002`, or
 `DYNAMO_IDX003`). Candidates whose partition key is not constrained by the query are not treated as
-rejections; scan-like base-table reads are reported by `ScanLikeQueryDetected` instead.
+rejections; reads that also fail to constrain the active source key may be reported by
+`ScanLikeQueryDetected` instead.
 Enable `Information` level logging for the `Query` category to see the full decision trail.
 
 ______________________________________________________________________
@@ -255,10 +256,11 @@ At least one secondary index was targeted by the predicate, but every targeted c
 later safety gate. The query will use the base table.
 
 This event is not emitted for unkeyed reads such as `ToListAsync()` or predicates that do not
-constrain any configured secondary-index partition key; those shapes are handled by
-`ScanLikeQueryDetected`. Review the preceding `DYNAMO_IDX005` rejection diagnostics to see why the
-targeted index candidates were rejected, then either adjust the query/index shape or confirm that
-running against the base table is acceptable for this access pattern.
+constrain any configured secondary-index partition key. Those shapes stay on the base table; when
+they also fail to constrain the active source key, `ScanLikeQueryDetected` may report the scan-like
+read. Review the preceding `DYNAMO_IDX005` rejection diagnostics to see why the targeted index
+candidates were rejected, then either adjust the query/index shape or confirm that running against
+the base table is acceptable for this access pattern.
 
 Use EF Core warning configuration to escalate or suppress this warning:
 
--- a/src/EntityFrameworkCore.DynamoDb/Query/Internal/DynamoAutoIndexSelectionAnalyzer.cs
+++ b/src/EntityFrameworkCore.DynamoDb/Query/Internal/DynamoAutoIndexSelectionAnalyzer.cs
@@ -328,6 +328,8 @@ internal sealed class DynamoAutoIndexSelectionAnalyzer : IDynamoIndexSelectionAn
     {
         var reason = gateResult switch
         {
+            // Unreachable from Analyze: NoPkConstraint is filtered there because scan
+            // classification owns user-facing warnings for non-keyed/base-table reads.
             CandidateGateResult.NoPkConstraint =>
                 "no equality or IN constraint on the index partition key",
             CandidateGateResult.UnsafeOr =>
--- a/src/EntityFrameworkCore.DynamoDb/Query/Internal/DynamoAutoIndexSelectionAnalyzer.cs
+++ b/src/EntityFrameworkCore.DynamoDb/Query/Internal/DynamoAutoIndexSelectionAnalyzer.cs
@@ -22,7 +22,7 @@ namespace EntityFrameworkCore.DynamoDb.Query.Internal;
 /// <para>
 /// Diagnostic codes:
 /// <list type="bullet">
-///   <item><c>DYNAMO_IDX001</c> — no candidate satisfies the predicate (Warning)</item>
+///   <item><c>DYNAMO_IDX001</c> — targeted candidates failed safety gates (Warning)</item>
 ///   <item><c>DYNAMO_IDX002</c> — multiple candidates tie (Warning)</item>
 ///   <item><c>DYNAMO_IDX003</c> — a single candidate was selected or would be auto-selected (Information)</item>
 ///   <item><c>DYNAMO_IDX004</c> — explicit index selected via .WithIndex() (Information)</item>
@@ -141,9 +141,9 @@ internal sealed class DynamoAutoIndexSelectionAnalyzer : IDynamoIndexSelectionAn
                 new(
                     DynamoQueryDiagnosticLevel.Warning,
                     "DYNAMO_IDX001",
-                    $"No secondary index on table '{context.SelectExpression.TableName}' satisfies "
-                    + "the predicate. The query will use the base table. "
-                    + "Ensure the WHERE clause includes an equality constraint on the index partition key."),
+                    $"No targeted secondary index on table '{context.SelectExpression.TableName}' "
+                    + "satisfies all safety gates. The query will use the base table. "
+                    + "Review preceding DYNAMO_IDX005 diagnostics for rejection reasons."),
             };
             return new DynamoIndexSelectionDecision(
                 null,
--- a/src/EntityFrameworkCore.DynamoDb/Query/Internal/DynamoAutoIndexSelectionAnalyzer.cs
+++ b/src/EntityFrameworkCore.DynamoDb/Query/Internal/DynamoAutoIndexSelectionAnalyzer.cs
@@ -107,11 +107,16 @@ internal sealed class DynamoAutoIndexSelectionAnalyzer : IDynamoIndexSelectionAn
             var gateResult = EvaluateGates(descriptor, constraints);
             if (gateResult != CandidateGateResult.Passed)
             {
-                rejectionDiagnostics.Add(
-                    MakeRejectionDiagnostic(
-                        descriptor,
-                        context.SelectExpression.TableName,
-                        gateResult));
+                // A missing partition-key constraint means the query did not target this index at
+                // all. Do not surface that as an index-selection problem; scan classification owns
+                // the user-facing warning for non-keyed/base-table reads.
+                if (gateResult != CandidateGateResult.NoPkConstraint)
+                    rejectionDiagnostics.Add(
+                        MakeRejectionDiagnostic(
+                            descriptor,
+                            context.SelectExpression.TableName,
+                            gateResult));
+
                 continue;
             }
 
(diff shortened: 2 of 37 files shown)


tests/EntityFrameworkCore.DynamoDb.Tests/Query/DynamoAutoIndexSelectionAnalyzerTests.cs
```diff
@@ -351,14 +351,14 @@ public void SuggestOnly_SingleMatch_EmitsInfoDiagnostic_Returns_NoSelection()
     /// <summary>Provides functionality for this member.</summary>
     [Fact(Timeout = TestConfiguration.DefaultTimeout)]
     /// <summary>Provides functionality for this member.</summary>
-    public void SuggestOnly_NoMatch_EmitsWarningDiagnostic_Returns_NoSelection()
+    public void SuggestOnly_NoMatch_Returns_NoSelectionWithoutIndexDiagnostics()
```
ncipollina, line 329: These `/// <summary>Provides functionality for this member.</summary>` comments appear on **every test method** throughout this file (75 occurrences). They are auto-generated IDE placeholder text and should be removed entirely. Per CLAUDE.md: "Do not add XML docs to tests, test helpers, local functions, or obvious private/internal members unless useful." The method names are already self-documenting — the doc comments add noise without value. A global find-and-replace removing the two-line pattern will clear them all at once.

tests/EntityFrameworkCore.DynamoDb.SpecificationTests/DynamoConcurrencyTest.cs
```diff
@@ -0,0 +1,213 @@
...
+                customer.Name = "Updated";
+                customer.Version = 2;
+                customer.LoyaltyLevel = "Silver";
+            },
+            async ctx => ctx.PremiumCustomers.Remove(
+                await ctx.PremiumCustomers.FirstAsync(c => c.Id == "5", CancellationToken.None)));
+
+    private Task ConcurrencyTestAsync<TException>(Func<ConcurrencyContext, Task> change)
+        where TException : DbUpdateException
+        => ConcurrencyTestAsync<TException, Customer>(null, change, change);
```
ncipollina, line 137: The single-delegate overload passes `change` as **both** `storeChange` and `clientChange`:

```csharp
=> ConcurrencyTestAsync<TException, Customer>(null, change, change);
```

In the core helper, `storeChange` runs on `innerContext` and `clientChange` runs on `outerContext`. With the same delegate for both, the test simulates: seed nothing, apply the same change on a fresh inner context (commit succeeds), then apply the same change again on `outerContext` (which never saw the inner context's write). This is used only by `Adding_the_same_entity_twice_results_in_DbUpdateException`, where the change `Add`s the same entity. After `innerContext.SaveChangesAsync` the entity exists in DynamoDB. When `outerContext.SaveChangesAsync` runs, `outerContext` still has that entity in `Added` state, which triggers a duplicate-key write — so the intent is correct. However the mechanics are subtle enough to warrant a comment explaining the setup, especially since the single-argument overload looks like it should only apply the change once but actually applies it twice on different contexts.

tests/EntityFrameworkCore.DynamoDb.SpecificationTests/DynamoConcurrencyTest.cs
```diff
@@ -0,0 +1,213 @@
...
+        Func<ConcurrencyContext, Task> clientChange) where TException : DbUpdateException
+    {
+        await using var outerContext = CreateContext();
+        await Fixture.TestStore.CleanAsync(outerContext);
+
+        if (seedAction != null)
+            await seedAction(outerContext);
+
+        await outerContext.SaveChangesAsync(CancellationToken.None);
+
```
ncipollina, line 160: The `storeChange` / `clientChange` role assignment in the harness appears inverted relative to standard EF Core concurrency test convention. In the helper, the parameter labelled `clientChange` runs on `outerContext` **before** `innerContext` commits, and `storeChange` runs on `innerContext` and commits first. In every call site, the second positional argument (maps to `storeChange`) is an update and the third (maps to `clientChange`) is the delete or second update. That means the inner context wins the race and the outer context's stale write is the one that should be rejected. This is correct behaviour — but the naming is backwards: conventionally `clientChange` is the write that races and loses, and `storeChange` is the write that already happened in the store. Swapping the parameter names (or reordering the call sites) would eliminate the confusion.

tests/EntityFrameworkCore.DynamoDb.SpecificationTests/DynamoConcurrencyTest.cs
```diff
@@ -0,0 +1,213 @@
+using EntityFrameworkCore.DynamoDb.SpecificationTests.TestUtilities;
+using Microsoft.EntityFrameworkCore;
+using Microsoft.EntityFrameworkCore.TestUtilities;
+
+namespace EntityFrameworkCore.DynamoDb.SpecificationTests;
+
+#nullable disable
```
ncipollina, line 7: `#nullable disable` suppresses nullable analysis for the whole file. The entity classes (`Customer`, `PremiumCustomer`) have mutable string properties (`Id`, `Name`, `LoyaltyLevel`) that are set only via object initializers, so they genuinely need initialization handling. Instead of blanket-disabling, consider either initializing them to `string.Empty` / `null!` with a nullable-enabled pragma, or restricting the disable to the entity class region. Blanket `#nullable disable` hides accidental nulls in the rest of the test logic (e.g., the async delegates).

tests/EntityFrameworkCore.DynamoDb.SpecificationTests/DynamoConcurrencyTest.cs
```diff
@@ -0,0 +1,213 @@
...
+
+[Collection(DynamoSpecificationCollection.Name)]
+public sealed class DynamoConcurrencyTest(DynamoConcurrencyTest.DynamoConcurrencyFixture fixture)
```
ncipollina, line 15: `DynamoConcurrencyTest` implements `IClassFixture` directly rather than extending one of the EF Core spec base classes (like `ConcurrencyTestBase`). That is intentional given that DynamoDB's optimistic concurrency semantics differ from relational providers. However, the pattern used by every other spec test in this project requires a `Check_all_tests_overridden` guard (see CLAUDE.md: "Every specification test class must include `Check_all_tests_overridden`"). Since this class does not inherit from a spec base, the guard doesn't apply — but there is no comment explaining why this class bypasses the shared infrastructure. A brief comment at the class level clarifying that this is a provider-custom concurrency test (not a port of a spec base) would prevent future contributors from questioning the missing guard.

tests/EntityFrameworkCore.DynamoDb.SpecificationTests/ConcurrencyDetectorEnabledDynamoTest.cs
```diff
@@ -0,0 +1,71 @@
...
+
+        await ConcurrencyDetectorTest(async c =>
+        {
+            c.Products.Add(new Product { Id = 2, Name = "Unicorn Replacement Horn Pack" });
+            return await c.SaveChangesAsync();
+        });
+
+        await using var ctx = CreateContext();
+        var newProduct = await ctx.Products.FirstOrDefaultAsync(p => p.Id == 2);
+        Assert.Null(newProduct);
```
ncipollina, line 51: After the `ConcurrencyDetectorTest` call in the `SaveChanges` (enabled) override, you assert `Assert.Null(newProduct)`. This implies that `SaveChangesAsync` under the enabled concurrency detector **does not persist** the product. That is the expected behavior for `ConcurrencyDetectorEnabledTestBase` — the detector is supposed to throw when a concurrent operation is detected. Double-check that this assertion is genuinely testing the detector behavior rather than a no-op: if `ConcurrencyDetectorTest` already asserts the exception internally, the post-check here is verifying a side effect of the throw, which is fine. But if the test can silently pass because the entity was never inserted (e.g., the fixture table is empty), the assertion would pass vacuously. A short comment noting what the null check is validating would clarify intent.

tests/EntityFrameworkCore.DynamoDb.SpecificationTests/TestUtilities/DynamoTestStore.cs
```diff
@@ -14,6 +14,15 @@ public class DynamoTestStore(string name, bool shared, DynamoTestStoreFactory fa
     public override DbContextOptionsBuilder AddProviderOptions(DbContextOptionsBuilder builder)
         => builder.UseDynamo(options => options.DynamoDbClient(Client));
 
+    /// <inheritdoc />
+    public override async Task CleanAsync(DbContext context)
+    {
+        context.ChangeTracker.Clear();
+
+        await context.Database.EnsureDeletedAsync().ConfigureAwait(false);
+        await context.Database.EnsureCreatedAsync().ConfigureAwait(false);
```
ncipollina, line 23: The project style (visible throughout the file and in the existing `InitializeAsync`) uses `.ConfigureAwait(false)` on library-context awaits. `EnsureCreatedAsync` is missing it here while `EnsureDeletedAsync` has it. For consistency with the existing `InitializeAsync` method and the project's library-code style, add `.ConfigureAwait(false)` to the `EnsureCreatedAsync` call as well.

docs/diagnostics.md
```diff
@@ -241,22 +241,26 @@ Index selection runs when [automatic index selection](querying/index-selection.m
...
 
-When multiple candidates are evaluated, `DYNAMO_IDX005` rejection events appear first, followed by
-the final outcome event (`DYNAMO_IDX001`, `DYNAMO_IDX002`, or `DYNAMO_IDX003`). Enable
-`Information` level logging for the `Query` category to see the full decision trail.
+When multiple targeted candidates are evaluated and rejected, `DYNAMO_IDX005` rejection events
+appear first, followed by the final outcome event (`DYNAMO_IDX001`, `DYNAMO_IDX002`, or
+`DYNAMO_IDX003`). Candidates whose partition key is not constrained by the query are not treated as
+rejections; reads that also fail to constrain the active source key may be reported by
+`ScanLikeQueryDetected` instead.
+Enable `Information` level logging for the `Query` category to see the full decision trail.
```
ncipollina, line 249: The update correctly adds context about when `DYNAMO_IDX001` is **not** emitted (queries that don't constrain any GSI partition key). However, the `DYNAMO_IDX005` section later in this file (line ~327, outside this diff) still lists "no equality or `IN` constraint on the index partition key" as a common rejection reason and shows it in an example log line. With `NoPkConstraint` now suppressed, that rejection reason is never actually emitted as IDX005. The IDX005 section's example and bullet should be updated in a follow-up to remove the stale case, to avoid confusing users who search their logs for that message.

src/EntityFrameworkCore.DynamoDb/Query/Internal/DynamoAutoIndexSelectionAnalyzer.cs
```diff
@@ -107,11 +107,16 @@ public DynamoIndexSelectionDecision Analyze(DynamoIndexAnalysisContext context)
...
+                // A missing partition-key constraint means the query did not target this index at
+                // all. Do not surface that as an index-selection problem; scan classification owns
+                // the user-facing warning for non-keyed/base-table reads.
+                if (gateResult != CandidateGateResult.NoPkConstraint)
+                    rejectionDiagnostics.Add(
+                        MakeRejectionDiagnostic(
+                            descriptor,
+                            context.SelectExpression.TableName,
+                            gateResult));
+
```
ncipollina, line 119: The guard is correct and the intent clear. One follow-up worth noting: `MakeRejectionDiagnostic` (outside this diff) still has a `NoPkConstraint` arm in its switch expression even though that case can no longer reach it. That dead arm is benign but could mislead a future reader into thinking IDX005 still fires for missing PK constraints. A comment on that arm (e.g. `// unreachable: NoPkConstraint is filtered above`) would prevent the confusion without requiring a behavior change.

src/EntityFrameworkCore.DynamoDb/Query/Internal/DynamoAutoIndexSelectionAnalyzer.cs
```diff
@@ -122,16 +127,23 @@ public DynamoIndexSelectionDecision Analyze(DynamoIndexAnalysisContext context)
         // ── 4. No usable candidate ───────────────────────────────────────────
         if (usableCandidates.Count == 0)
         {
+            if (rejectionDiagnostics.Count == 0)
+                return new DynamoIndexSelectionDecision(
+                    null,
+                    DynamoIndexSelectionReason.NoSelection,
+                    []);
```
ncipollina, line 134: The early-exit when `rejectionDiagnostics.Count == 0` is correct for the intended semantics. One edge case worth verifying: if a table has **no secondary indexes at all** (i.e., `context.CandidateDescriptors` contains only the base-table descriptor), the `foreach` loop body is never entered, `usableCandidates` and `rejectionDiagnostics` are both empty, and the code falls through to this branch and returns `NoSelection` with empty diagnostics. That is the right outcome — there is nothing to select. But the `IsFullBaseTableKeyLookup` short-circuit above it only triggers when the base-table PK is fully constrained. A no-index table with an unkeyed query would also reach this path, which is fine. This is not a bug — just worth a unit test covering the zero-index case to lock it down.

ncipollina (commented):
## Review summary

Overall the PR is a solid increment. The analyzer narrowing is correct, the `CleanAsync` centralization is a good DRY win, and the concurrency test coverage fills a real gap. Below are the items flagged inline, ordered by priority.

### Must fix before merge

**Stale XML doc comments — `DynamoAutoIndexSelectionAnalyzerTests.cs`**
75 occurrences of `/// <summary>Provides functionality for this member.</summary>` appear on test methods across the file. Per CLAUDE.md these should not exist on test methods. A single global find-and-replace removes them all. The pattern to delete is the two-line block `/// <summary>Provides functionality for this member.</summary>\n    /// <summary>Provides functionality for this member.</summary>` (one appears before `[Fact]` and one between the attribute and the method).

**`DYNAMO_IDX005` docs still list the suppressed rejection reason**
`docs/diagnostics.md` lines ~326-334 (outside the PR diff) still show "no equality or IN constraint on the index partition key" as a common IDX005 rejection reason with an example log line. That message is now never emitted because `NoPkConstraint` is suppressed. The bullet and example should be removed or replaced with a note that this case is handled silently.

### Should fix

**`ConcurrencyTestAsync` parameter naming is inverted**
`storeChange` runs on `innerContext` (commits first, wins the race) and `clientChange` runs on `outerContext` (the stale context whose save is expected to fail). By convention `storeChange` is the write already committed to the store and `clientChange` is the stale in-flight write. The names as written imply the opposite assignment. Renaming the parameters (or flipping the call-site argument order with a matching rename) removes the cognitive overhead.

**`#nullable disable` covers the whole file in `DynamoConcurrencyTest.cs` and `ConcurrencyDetectorEnabledDynamoTest.cs`**
The disable is only needed for the entity classes with uninitialized string properties. Narrowing it to those classes (or initializing the properties) keeps nullability analysis active for the test logic.

**`CleanAsync` — missing `.ConfigureAwait(false)` on `EnsureCreatedAsync`**
`EnsureDeletedAsync` has it, `EnsureCreatedAsync` does not. One-line fix.

### Nice to have / follow-up

- `MakeRejectionDiagnostic` in the analyzer has an unreachable `NoPkConstraint` arm. It's benign but a `// unreachable: NoPkConstraint is filtered above` comment prevents future confusion.
- The single-delegate `ConcurrencyTestAsync<TException>` overload passes `change` as both `storeChange` and `clientChange`. The resulting behavior (duplicate-key insert via two contexts) is intentional for `Adding_the_same_entity_twice`, but a brief comment explaining the two-context setup would make the mechanics obvious.
- A unit test covering the zero-secondary-index case (base-table descriptor only, unkeyed query) would lock down the new early-exit path.
- `DynamoConcurrencyTest` does not inherit from a spec base and has no `Check_all_tests_overridden` guard — a class-level comment explaining it is a provider-custom test (not a spec port) would preempt the obvious question.
- The `ConcurrencyDetectorEnabledDynamoTest.SaveChanges` null assertion would benefit from a comment explaining what "null" validates (the save threw, so nothing was persisted).

ncipollina (changes requested):
⚠️ Left some feedback.

tests/EntityFrameworkCore.DynamoDb.Tests/Query/DynamoAutoIndexSelectionAnalyzerTests.cs
j-d-ha: Fixed!

tests/EntityFrameworkCore.DynamoDb.SpecificationTests/DynamoConcurrencyTest.cs
j-d-ha: Fixed!

tests/EntityFrameworkCore.DynamoDb.SpecificationTests/DynamoConcurrencyTest.cs
j-d-ha: Fixed!

tests/EntityFrameworkCore.DynamoDb.SpecificationTests/DynamoConcurrencyTest.cs
j-d-ha: Fixed!

tests/EntityFrameworkCore.DynamoDb.SpecificationTests/DynamoConcurrencyTest.cs
j-d-ha: Fixed!

tests/EntityFrameworkCore.DynamoDb.SpecificationTests/ConcurrencyDetectorEnabledDynamoTest.cs
j-d-ha: I don't think that this one is real because we are actually fully writing out the test in this case, because the base test case uses single, which we cannot support.

tests/EntityFrameworkCore.DynamoDb.SpecificationTests/TestUtilities/DynamoTestStore.cs
j-d-ha: Already fixed.

docs/diagnostics.md
j-d-ha: Fixed!

src/EntityFrameworkCore.DynamoDb/Query/Internal/DynamoAutoIndexSelectionAnalyzer.cs
j-d-ha: Fixed!

src/EntityFrameworkCore.DynamoDb/Query/Internal/DynamoAutoIndexSelectionAnalyzer.cs
j-d-ha: Fixed!

ncipollina (approved):
🚢 All review feedback addressed — this is ready to merge.

**What was fixed:**
- Stale `/// <summary>Provides functionality for this member.</summary>` comments scrubbed across all test files
- IDX005 docs section updated to remove the now-suppressed `NoPkConstraint` rejection example and replace it with the correct `unsafeOR` example
- `storeChange`/`clientChange` renamed to `innerContextChange`/`outerContextChange` — intent is now unambiguous
- `#nullable disable` retained (entity strings genuinely need it) with a class-level doc explaining why this is a provider-custom test
- `.ConfigureAwait(false)` now consistent on both `EnsureDeletedAsync` and `EnsureCreatedAsync` in `DynamoTestStore`
- `// Unreachable from Analyze: NoPkConstraint is filtered there...` comment added to the dead arm in `MakeRejectionDiagnostic`
- `RejectionDiagnostics_PrecedeIdx003` test correctly narrowed to only emit IDX003

Logic is correct, diagnostics contract is clean, test coverage is solid. Ship it. ✅

PR #231, opened 2026-05-19
test(tests): add Northwind where specification coverage

ncipollina:
## Summary

Adds DynamoDB specification coverage for EF Core's Northwind where-query test suite. The new fixture maps the Northwind model to DynamoDB tables, adapts relationship/key behavior for the provider, and explicitly skips query shapes DynamoDB does not support.

## Changes

- Added `NorthwindWhereQueryDynamoTest` with full overrides for the EF Core `NorthwindWhereQueryTestBase` surface.
- Added `NorthwindQueryDynamoFixture` for Northwind table mappings, seed execution, scan-like query handling, and DynamoDB transaction chunking during large seed operations.
- Extended spec test helpers with async-aware sync-query validation.
- Adjusted provider model/convention validation so spec fixtures can reshape EF models before final DynamoDB validation.
- Fixed XML docs for extension-block query APIs so the solution builds cleanly.
- Included existing working-tree documentation/test cleanup changes per current workspace state.

## Validation

- `dotnet build`
- `dotnet test --project tests/EntityFrameworkCore.DynamoDb.SpecificationTests/EntityFrameworkCore.DynamoDb.SpecificationTests.csproj --no-build --filter "FullyQualifiedName~NorthwindWhereQueryDynamoTestDefault"`
  - 193 succeeded
  - 107 skipped
  - 0 failed
- `dotnet test --project tests/EntityFrameworkCore.DynamoDb.SpecificationTests/EntityFrameworkCore.DynamoDb.SpecificationTests.csproj --no-build --filter "FullyQualifiedName~NorthwindWhereQueryDynamoTestDefault.Check_all_tests_overridden"`

## Notes for Reviewers

The skipped Northwind cases are unsupported DynamoDB query shapes such as navigation/subquery predicates, row-count operators (`Take`/`Skip` semantics), and other translations that do not map cleanly to PartiQL/DynamoDB behavior.


src/EntityFrameworkCore.DynamoDb/Metadata/Conventions/DynamoRelationshipValidationConvention.cs
```diff
@@ -7,14 +7,8 @@ namespace EntityFrameworkCore.DynamoDb.Metadata.Conventions;
...
 public sealed class DynamoRelationshipValidationConvention
-    : IForeignKeyAddedConvention, IForeignKeyOwnershipChangedConvention, IModelFinalizingConvention
+    : IForeignKeyOwnershipChangedConvention, IModelFinalizingConvention
 {
-    /// <inheritdoc />
-    public void ProcessForeignKeyAdded(
-        IConventionForeignKeyBuilder foreignKeyBuilder,
-        IConventionContext<IConventionForeignKeyBuilder> context)
-        => ThrowIfUnsupportedRelationship(foreignKeyBuilder.Metadata);
-
```
j-d-ha, line 17: Why was this change needed?

tests/EntityFrameworkCore.DynamoDb.SpecificationTests/Query/NorthwindQueryDynamoFixture.cs
```diff
@@ -0,0 +1,79 @@
...
+        modelBuilder.Ignore<CustomerQuery>();
+        modelBuilder.Ignore<CustomerQueryWithQueryFilter>();
+        modelBuilder.Ignore<OrderQuery>();
+        modelBuilder.Ignore<ProductQuery>();
+        modelBuilder.Ignore<ProductView>();
+
+        modelBuilder.Entity<Customer>().ToTable("Customers").HasPartitionKey(e => e.CustomerID);
+        modelBuilder.Entity<Employee>().ToTable("Employees").HasPartitionKey(e => e.EmployeeID);
+        modelBuilder.Entity<Order>().ToTable("Orders").HasPartitionKey(e => e.OrderID);
+        modelBuilder.Entity<Product>().ToTable("Products").HasPartitionKey(e => e.ProductID);
```
j-d-ha, line 67: You probably don't need to say the table name as a default to the entity name.

tests/EntityFrameworkCore.DynamoDb.SpecificationTests/Query/NorthwindWhereQueryDynamoTest.cs
```diff
@@ -0,0 +1,1714 @@
...
+            await base.Where_simple_shadow_projection(a);
+            AssertSql(
+            """
+            SELECT "title"
+            FROM "Employees"
+            WHERE "title" = 'Sales Representative'
+            """);
+        });
+
+    [ConditionalTheory(Skip = "DynamoDB provider does not support this query shape.")]
```
j-d-ha: Can we add the skip reason to the file with the other skip reasons in the root of the specification test project?

tests/EntityFrameworkCore.DynamoDb.SpecificationTests/Query/NorthwindWhereQueryDynamoTest.cs
```diff
@@ -0,0 +1,1714 @@
...
+    public override Task Where_simple_shadow_projection_mixed(bool async)
+        => Task.CompletedTask;
+
+    [ConditionalTheory(Skip = "DynamoDB provider does not support this query shape.")]
+    public override Task Where_simple_shadow_subquery(bool async)
+        => Task.CompletedTask;
+
+    [ConditionalTheory(Skip = "DynamoDB provider does not support this query shape.")]
+    public override Task Where_shadow_subquery_FirstOrDefault(bool async)
```
j-d-ha, line 412: Do you think it's worth writing a dynamo compatible version of this test?

j-d-ha (changes requested):
Looks really good. Left a few small requests.

src/EntityFrameworkCore.DynamoDb/Metadata/Conventions/DynamoRelationshipValidationConvention.cs
ncipollina: The change was needed because IForeignKeyAddedConvention validates during intermediate model construction, before provider fixtures/customization can remove unsupported relationships. The Northwind spec model starts with relational navigations/FKs that our DynamoDB fixture ignores. Immediate validation caused those transient relationships to fail too early. By validating at model finalization instead, we still reject any real FK relationships left in the final DynamoDB model, while allowing temporary relationships created by shared EF test models to be removed first.

tests/EntityFrameworkCore.DynamoDb.SpecificationTests/Query/NorthwindQueryDynamoFixture.cs
ncipollina: It’s not technically required. Without `ToTable(...)`, the provider falls back to the CLR type name (`Customer`, `Employee`, `Order`, `Product`) rather than the Northwind logical table/`DbSet` names (`Customers`, `Employees`, `Orders`, `Products`).

I used explicit `ToTable(...)` calls to preserve the canonical Northwind table names in the generated PartiQL baselines, e.g.:

```sql
FROM "Customers"
```

instead of:

```sql
FROM "Customer"
```

The tests would likely still work if we removed those calls and updated the baselines, because the test store is created/seeded from the model. But keeping them makes this fixture line up better with the Northwind model and with other provider baselines.

So I’d say it is intentional for consistency/readability rather than strictly required for correctness. If we prefer less explicit configuration and are okay with singular provider-default table names in the baselines, we can remove the `ToTable(...)` calls and update the expected PartiQL.

tests/EntityFrameworkCore.DynamoDb.SpecificationTests/Query/NorthwindWhereQueryDynamoTest.cs
ncipollina: Done.

tests/EntityFrameworkCore.DynamoDb.SpecificationTests/Query/NorthwindWhereQueryDynamoTest.cs
ncipollina: I don’t think it’s worth adding a Dynamo-compatible rewrite for this specific spec test in this PR.

The shape being skipped is the nested scalar subquery with `FirstOrDefault()` inside the predicate, not just `EF.Property`/shadow-property access:

```csharp
from e in ss.Set<Employee>()
where EF.Property<string>(e, "Title")
    == EF.Property<string>(
        ss.Set<Employee>()
            .OrderBy(e2 => EF.Property<string>(e2, "Title"))
            .FirstOrDefault(),
        "Title")
select e
```

A Dynamo-compatible version would probably have to split this into two separate queries:

```csharp
var title = await context.Employees
    .OrderBy(e => e.Title)
    .Select(e => e.Title)
    .FirstOrDefaultAsync();

var employees = await context.Employees
    .Where(e => EF.Property<string>(e, "Title") == title)
    .ToListAsync();
```

But that is no longer testing the same translation behavior as the EF spec test. It tests client-side composition across two queries, not translation of a scalar subquery in a predicate.

So I’d keep this skipped for now and add targeted coverage later if/when we add scalar subquery translation support.

src/EntityFrameworkCore.DynamoDb/Metadata/Conventions/DynamoRelationshipValidationConvention.cs
j-d-ha: I assume we have tests that cover this case?

j-d-ha (approved):
👍 LGTM!

PR #232, opened 2026-05-19
test(tests): add Northwind select specification coverage

ncipollina:
## Summary

Adds DynamoDB EF Core specification coverage for the shared Northwind select-query test suite. The new provider test class overrides every inherited `NorthwindSelectQueryTestBase` case, runs supported shapes against DynamoDB Local with PartiQL baselines, and explicitly skips unsupported query shapes using centralized skip reasons.

## Changes

- Added `NorthwindSelectQueryDynamoTest` under `tests/EntityFrameworkCore.DynamoDb.SpecificationTests/Query/`.
- Added override coverage for all inherited Northwind select spec tests with the same guard pattern used by the where-query specs.
- Captured PartiQL baselines for supported select projections.
- Marked unsupported DynamoDB query shapes as skipped, including joins/`SelectMany`, navigations/correlated collections, `Distinct`/grouping shapes, unrestricted ordering/reverse shapes, nested subqueries, and unsupported object/collection projection forms.
- Added a more precise `SkipReason.OrderedResultSetNotSupported` for select specs whose real blocker is guaranteed ordering rather than cast/projection translation.
- Fixed client-only select projections (`Select(c => new { })`, constants, `new DateTime()`) by emitting a minimal explicit attribute projection so DynamoDB can return one row per matching item.
- Documented client-only projection behavior in `docs/querying/projection.md`.

## Validation

- `dotnet build`
- `dotnet test --project tests/EntityFrameworkCore.DynamoDb.SpecificationTests/EntityFrameworkCore.DynamoDb.SpecificationTests.csproj --no-build --filter "FullyQualifiedName~NorthwindSelectQueryDynamoTestDefault" -- --no-progress`
- `uv run zensical build`

Select spec result:

```text
total: 238
failed: 0
succeeded: 103
skipped: 135
```

## Notes for Reviewers

This follows the pattern from the Northwind where-query spec work: start with the full inherited surface, keep supported cases asserting generated PartiQL, and classify currently unsupported shapes with explicit skip reasons rather than silently omitting them.

### DIFF
--- a/docs/querying/projection.md
+++ b/docs/querying/projection.md
@@ -57,6 +57,10 @@ var results = await db.Products
 
 This means the server always fetches the underlying attributes; there is no server-side function evaluation for computed projection expressions. Plan your attribute reads accordingly — if a projection reads several attributes to compute one value, all of those attributes are fetched from DynamoDB.
 
+For client-only projections that do not read any entity attributes, such as constants or `new { }`,
+the provider still emits a minimal explicit attribute projection so DynamoDB returns one row per
+matching item. The client-side shaper then constructs the requested constant value for each row.
+
 ## Primitive Collection Properties
 
 Entity properties typed as collections materialize from DynamoDB set and list attribute types:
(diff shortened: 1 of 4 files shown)


j-d-ha (approved):
👍 LGTM!

PR #233, opened 2026-05-20
test(tests): add Northwind aggregate specification coverage

ncipollina:
## Summary

Adds DynamoDB coverage for EF Core's `NorthwindAggregateOperatorsQueryTestBase`. The new specification test class mirrors the existing Northwind where/select pattern: supported aggregate/operator cases execute against the shared DynamoDB Northwind fixture with PartiQL baselines, while unsupported query shapes are explicitly skipped with the centralized skip reason.

## Changes

- Added `NorthwindAggregateOperatorsQueryDynamoTest` with overrides for all inherited aggregate-operator specification tests.
- Captured PartiQL baselines for supported cases using `Fixture.AssertSql(...)`.
- Marked unsupported aggregate/query shapes with `SkipReason.QueryShapeNotSupported` so current provider limits are explicit.

## Validation

- `dotnet build`
- `dotnet test --project tests/EntityFrameworkCore.DynamoDb.SpecificationTests/EntityFrameworkCore.DynamoDb.SpecificationTests.csproj --no-build --filter "FullyQualifiedName~NorthwindAggregateOperatorsQueryDynamoTestDefault" -- --no-progress`
  - `266 total`
  - `109 succeeded`
  - `157 skipped`
  - `0 failed`

## Notes for Reviewers

This PR is test-only. The skipped cases reflect currently unsupported DynamoDB provider query shapes discovered by first running the full inherited suite unskipped, then classifying failures against existing provider capabilities.

### DIFF
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -8,28 +8,28 @@
     <PackageVersion Include="AutoFixture.AutoNSubstitute" Version="4.18.1" />
     <PackageVersion Include="AutoFixture.Xunit3" Version="4.19.0" />
     <PackageVersion Include="AwesomeAssertions" Version="9.4.0" />
-    <PackageVersion Include="AWSSDK.DynamoDBv2" Version="4.0.17.9" />
+    <PackageVersion Include="AWSSDK.DynamoDBv2" Version="4.0.18.4" />
     <PackageVersion Include="Humanizer.Core" Version="3.0.10" />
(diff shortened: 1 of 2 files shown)


j-d-ha (changes requested):
I'm not sure if we should even be overriding this test, given the fact that we don't actually support any aggregation operations for Dynamo.

PR #245, opened 2026-05-21
test(tests): add complex type query specs

ncipollina:
## Summary

Adds DynamoDB provider specification coverage for `ComplexTypeQueryTestBase` and updates the medium-term spec-test plan. The implementation focuses on DynamoDB-aligned complex property projection/filtering while explicitly skipping navigation, set-operation, GroupBy, equality, subquery, and pushdown-heavy shapes that are not currently supported.

## Changes

- Added `ComplexTypeQueryDynamoTest` with explicit overrides for all inherited complex type query spec methods.
- Added a DynamoDB-specific `ComplexTypeQueryDynamoFixture` that maps/seeds `Customer` and `ValuedCustomer` with nested complex properties and ignores navigation-heavy group entities.
- Registered `ComplexTypeQueryTestBase<>` in `ComplianceDynamoTest`.
- Added shared skip reasons for set operations and GroupBy.
- Updated `docs/spec-test-coverage.md`:
  - moved `ComplexTypeQueryTestBase` to Implemented under Other Query
  - moved `InheritanceQueryDynamoTest` to Long-term with blockers (`OfType`, discriminator expression translation, fixture work)
  - refreshed implementation-order and totals

## Validation

- `dotnet test --project tests/EntityFrameworkCore.DynamoDb.SpecificationTests/EntityFrameworkCore.DynamoDb.SpecificationTests.csproj --no-restore --filter "FullyQualifiedName~ComplexTypeQueryDynamoTest|FullyQualifiedName~ComplianceDynamoTest"`
  - Passed: 84 total, 18 succeeded, 66 skipped, 0 failed


tests/EntityFrameworkCore.DynamoDb.SpecificationTests/Query/ComplexTypeQueryDynamoTest.cs
```diff
@@ -0,0 +1,454 @@
...
+            WHERE "shippingAddress"."zipCode" = 7728
+            """);
+        });
+
+    [ConditionalTheory(Skip = SkipReason.SetOperationsNotSupported)]
+    public override Task Select_complex_type_Distinct(bool async)
+        => base.Select_complex_type_Distinct(async);
+
+    [ConditionalTheory(Skip = SkipReason.QueryShapeNotSupported)]
+    public override Task Complex_type_equals_complex_type(bool async)
```
j-d-ha, line 100: I think we might be able to support this one?

j-d-ha, line 100: Same here.

tests/EntityFrameworkCore.DynamoDb.SpecificationTests/Query/ComplexTypeQueryDynamoTest.cs
```diff
@@ -0,0 +1,454 @@
...
+        => base.Complex_type_equals_complex_type(async);
+
+    [ConditionalTheory(Skip = SkipReason.QueryShapeNotSupported)]
+    public override Task Complex_type_equals_constant(bool async)
+        => base.Complex_type_equals_constant(async);
+
+    [ConditionalTheory(Skip = SkipReason.QueryShapeNotSupported)]
+    public override Task Complex_type_equals_parameter(bool async)
```
j-d-ha, line 124: Same here

tests/EntityFrameworkCore.DynamoDb.SpecificationTests/Query/ComplexTypeQueryDynamoTest.cs
```diff
@@ -0,0 +1,454 @@
...
+            await base.Select_struct_complex_type(a);
+            AssertSql(
+            """
+            SELECT "shippingAddress"
+            FROM "ValuedCustomers"
+            """);
+        });
+
+    [ConditionalTheory(Skip = SkipReason.StructComplexTypeProjectionNotSupported)]
+    public override Task Select_nested_struct_complex_type(bool async)
```
j-d-ha, line 243: Same here

tests/EntityFrameworkCore.DynamoDb.SpecificationTests/Query/ComplexTypeQueryDynamoTest.cs
```diff
@@ -0,0 +1,454 @@
...
+        => base.Select_nested_struct_complex_type(async);
+
+    [ConditionalTheory(Skip = SkipReason.StructComplexTypeProjectionNotSupported)]
+    public override Task Select_single_property_on_nested_struct_complex_type(bool async)
```
j-d-ha, line 254: Same here

tests/EntityFrameworkCore.DynamoDb.SpecificationTests/Query/ComplexTypeQueryDynamoTest.cs
```diff
@@ -0,0 +1,454 @@
...
+            WHERE "shippingAddress"."zipCode" = 7728
+            """);
+        });
+
+    [ConditionalTheory(Skip = SkipReason.SetOperationsNotSupported)]
+    public override Task Select_struct_complex_type_Distinct(bool async)
+        => base.Select_struct_complex_type_Distinct(async);
+
+    [ConditionalTheory(Skip = SkipReason.QueryShapeNotSupported)]
+    public override Task Struct_complex_type_equals_struct_complex_type(bool async)
```
j-d-ha, line 281: Same here

j-d-ha (changes requested):
I think a couple of these skips might not be needed. Can we take another pass here? My latest PR might also unblock a few items.

Given that we only use complex types, I do really want to make sure we are dotting our I's and crossing out T's.

src/EntityFrameworkCore.DynamoDb/Query/Internal/DynamoShapedQueryCompilingExpressionVisitor.cs
j-d-ha: Nice Fix!

docs/spec-test-coverage.md
```diff
@@ -161,11 +161,16 @@ No Northwind query specification test classes are currently queued here.
 
 These tests use non-Northwind models and fixtures.
 
+### Implemented
+
+| Test Class | Methods | Cosmos | MongoDB | Notes |
+|---|---:|:---:|:---:|---|
+| `ComplexTypeQueryTestBase` | 74 | ✓ | ✗ | Complex property projection/filter, nested struct projections, and property-to-property/parameter equality coverage; navigation, set-operation, GroupBy, inline constant equality, and pushdown cases skipped |
```
j-d-ha: This row may overstate executed coverage. The targeted run reports many skipped cases (e.g. ~29 succeeded / 60 skipped for `ComplexTypeQueryDynamoTest`), while this lists `ComplexTypeQueryTestBase` under Implemented with 74 methods. Suggest wording this as registered/overridden with supported subsets called out, or splitting total inherited methods from executed behavioral coverage.

docs/modeling/complex-types.md
```diff
@@ -137,6 +137,21 @@ context.Customers.Where(x => x.Tags[0] == "featured")
...
+
+```csharp
+context.Customers.Where(x => x.ShippingAddress == x.BillingAddress)
+```
+
+```sql
+WHERE "shippingAddress" = "billingAddress"
+```
+
+Equality against a complex object parameter also translates and binds the value as a DynamoDB map
```
j-d-ha: Worth documenting two DynamoDB-specific caveats here: (1) complex map equality without a partition-key predicate is scan-like, and (2) whole-complex equality compares the exact stored DynamoDB map shape, so omitted nested attributes may not compare equal to explicit `NULL` entries.

src/EntityFrameworkCore.DynamoDb/Storage/DynamoComplexTypeMapping.cs
```diff
@@ -0,0 +1,34 @@
...
+    internal override bool RequiresParameterForPartiQlLiteral => true;
+
+    internal override AttributeValue CreateAttributeValue(object? value)
+        => EntityWritePlan.SerializeComplexTypeValue(value, complexType);
+
+    // sourceType is intentionally unused; the EF complex type metadata drives map serialization.
+    internal override AttributeValue CreateAttributeValue(object? value, Type sourceType)
+        => EntityWritePlan.SerializeComplexTypeValue(value, complexType);
+
+    public override string GenerateConstant(object? value)
```
j-d-ha, line 24: Inline complex constants still appear able to reach SQL generation and throw here. That matches the unsupported behavior, but the failure shape may be less clean than other translation rejections. Consider rejecting complex constants earlier in translation with provider error details and adding a targeted test for the failure mode.

src/EntityFrameworkCore.DynamoDb/Query/Internal/DynamoShapedQueryCompilingExpressionVisitor.cs
```diff
@@ -202,6 +206,39 @@ public override void AddStructuralTypeInitialization(
         }
     }
 
+    /// <summary>
+    ///     Rewrites value-type member access over try expressions into a temporary assignment.
```
j-d-ha: Comment says this rewrites member access over `TryExpression`, but `RequiresValueTypeInstanceMaterialization` below rewrites any non-parameter/member/constant value-type expression. Either narrow the predicate to the intended expression shapes or update this comment to describe the broader workaround.

j-d-ha (commented):
Adversarial review pass: leaving focused notes from parallel reviewers. No approval/blocking state implied.

j-d-ha (commented):
A few more comments

docs/spec-test-coverage.md
ncipollina: Updated. The row now says the 74 inherited methods are registered/overridden with explicit outcomes, and distinguishes the supported executed subsets from explicitly skipped cases.

docs/modeling/complex-types.md
ncipollina: Updated. Added DynamoDB-specific caveats for whole-complex equality:                                                                                                                                         
 - without a partition-key predicate it can be scan-like                                                                                                                                                      
 - equality compares the exact stored map shape, so omitted nested attributes may differ from explicit NULL

src/EntityFrameworkCore.DynamoDb/Storage/DynamoComplexTypeMapping.cs
ncipollina: Updated the error message to be clearer if inline complex constants ever reach SQL generation:                                                                                                               
                                                                                                                                                                                                              
 │ Inline complex type constants are not supported in DynamoDB queries. Assign the value to a variable so it can be sent as a query parameter.                                                                
                                                                                                                                                                                                              
 I did not add a separate targeted test in this PR because the spec skip already documents inline constant equality as unsupported, but the failure path is now clearer.

src/EntityFrameworkCore.DynamoDb/Query/Internal/DynamoShapedQueryCompilingExpressionVisitor.cs
ncipollina: Updated the comment to match the broader predicate. It now describes rewriting member access over non-trivial value-type expressions, including TryExpression from complex struct materialization.

tests/EntityFrameworkCore.DynamoDb.SpecificationTests/Query/ComplexTypeQueryDynamoTest.cs
```diff
@@ -0,0 +1,502 @@
...
+            AssertSql(
+            """
+            SELECT "id", "name", "billingAddress", "optionalAddress", "shippingAddress"
+            FROM "Customers"
+            WHERE "shippingAddress" = "billingAddress"
+            """);
+        });
+
+    [ConditionalTheory(Skip = SkipReason.ComplexTypeConstantEqualityNotSupported)]
+    public override Task Complex_type_equals_constant(bool async)
```
j-d-ha, line 112: I think this can be supported. This is the base test:

```csharp
    [ConditionalTheory, MemberData(nameof(IsAsyncData))]
    public virtual Task Complex_type_equals_constant(bool async)
        => AssertQuery(
            async,
            ss => ss.Set<Customer>().Where(c => c.ShippingAddress
                == new Address
                {
                    AddressLine1 = "804 S. Lakeshore Road",
                    ZipCode = 38654,
                    Country = new Country { FullName = "United States", Code = "US" },
                    Tags = new List<string> { "foo", "bar" }
                }),
            ss => ss.Set<Customer>().Where(c =>
                c.ShippingAddress.AddressLine1 == "804 S. Lakeshore Road"
                && c.ShippingAddress.ZipCode == 38654
                && c.ShippingAddress.Country == new Country { FullName = "United States", Code = "US" }
                && c.ShippingAddress.Tags.SequenceEqual(new List<string> { "foo", "bar" })));
```

tests/EntityFrameworkCore.DynamoDb.SpecificationTests/Query/ComplexTypeQueryDynamoTest.cs
```diff
@@ -0,0 +1,502 @@
...
+            AssertSql(
+            """
+            SELECT "id", "name", "billingAddress", "shippingAddress"
+            FROM "ValuedCustomers"
+            WHERE "shippingAddress" = "billingAddress"
+            """);
+        });
+
+    [ConditionalTheory(Skip = SkipReason.ComplexTypeConstantEqualityNotSupported)]
+    public override Task Struct_complex_type_equals_constant(bool async)
```
j-d-ha, line 293: This should also be supported. Base test case: 

```csharp
    [ConditionalTheory, MemberData(nameof(IsAsyncData))]
    public virtual Task Struct_complex_type_equals_constant(bool async)
        => AssertQuery(
            async,
            ss => ss.Set<ValuedCustomer>().Where(c => c.ShippingAddress
                == new AddressStruct
                {
                    AddressLine1 = "804 S. Lakeshore Road",
                    ZipCode = 38654,
                    Country = new CountryStruct { FullName = "United States", Code = "US" }
                }));
```

j-d-ha (changes requested):
Two more test cases, I think we need to support these.

tests/EntityFrameworkCore.DynamoDb.SpecificationTests/Query/ComplexTypeQueryDynamoTest.cs
ncipollina: Implemented.

tests/EntityFrameworkCore.DynamoDb.SpecificationTests/Query/ComplexTypeQueryDynamoTest.cs
ncipollina: Implemented.

j-d-ha (commented):
Thanks for the PR. A few things worth tightening before merge:

1. **Compliance registration gap**

`CustomConvertersDynamoTest` derives from `CustomConvertersTestBase<>`, but `ComplianceDynamoTest.GetBaseTestClasses()` does not include `CustomConvertersTestBase<>` (`tests/EntityFrameworkCore.DynamoDb.SpecificationTests/ComplianceDynamoTest.cs`). Please add `yield return typeof(CustomConvertersTestBase<>);` so future upstream custom-converter spec tests cannot be missed silently.

2. **Complex equality parity for `.Equals` / `object.Equals`**

Complex equality works for `==`, but `object.Equals(...)` / instance `.Equals(...)` do not appear to share the same complex-path fallback. `VisitBinary` has fallback for whole complex operands, but `TryTranslateEquals` translates operands directly and can reject whole complex properties / `DynamoComplexTypeMapping` (`DynamoSqlTranslatingExpressionVisitor.cs`, around `TryTranslateEquals` and `IsSupportedEqualsOperand`). Please add parity tests for shapes like `object.Equals(c.ShippingAddress, address)` and `c.ShippingAddress.Equals(address)`, then route `.Equals` through the same complex equality handling as `==`.

3. **Provider translator should not execute arbitrary user code**

`TryTranslateComplexConstant` currently compiles/invokes expression trees during query translation (`Expression.Lambda(...).Compile().Invoke()`). That can execute user code for shapes such as `e.ShippingAddress == BuildAddress()`. Please restrict this to EF-extracted constants/parameters or conservative literal-only shapes, or use EF's evaluatable-expression infrastructure; the provider translator should not execute arbitrary expression trees.

Optional but worth clarifying: complex equality currently translates as whole-map equality (`"shippingAddress" = "billingAddress"` / `?`). That can differ from EF relational structural equality if DynamoDB documents contain extra unmapped attributes. If whole-document equality is intentional for DynamoDB, please document that caveat.

j-d-ha (approved):
LGTM

PR #246, opened 2026-05-21
feat(efcore): improve value converter query support

j-d-ha:
## Summary

Adds broader DynamoDB query/type-mapping support around value converters, conditional expressions, member/property access, and primitive collection `Contains` translation. Expands spec, unit, and integration coverage for custom converters and primitive collection query behavior, and updates user-facing docs/limitations to match supported behavior.

## Changes

- Query translation
  - Adds PartiQL `contains` translation for primitive collections.
  - Improves conditional expression translation and translated member/property access.
  - Refines type mapping inference for custom struct conversions and converted values.
- Serialization and validation
  - Improves null model value handling for value converters.
  - Adds provider type validation for value converters.
  - Refines owned entity validation and DynamoDB read/write value handling.
- Tests
  - Adds custom converter specification coverage.
  - Adds primitive collection `Contains` translation and integration coverage.
  - Marks currently unsupported query shapes with explicit skips.
- Docs
  - Updates querying docs and limitations for `contains`, value conversion, and collection membership behavior.

## Validation

- Not run in this session.
- Added/updated unit, specification, and integration tests covering changed query and conversion behavior.

## Release Notes

- Adds support for primitive collection `Contains` translation to DynamoDB PartiQL `contains`.
- Improves value-converter support across query translation, type mapping, and serialization.


ncipollina (changes requested):
🔍 **Review: 2 blocking issues, 2 suggestions, 2 nits**

Solid feature work overall — the `contains` translation and value-converter improvements are well-structured. Two blocking issues need resolution before merge: a stack-balance bug in the new projection visitor, and the spec test class not following the `abstract` + sealed-inner-class pattern required by the shared container infrastructure.

See inline comments for details.

tests/EntityFrameworkCore.DynamoDb.SpecificationTests/CustomConvertersDynamoTest.cs
```diff
@@ -0,0 +1,710 @@
...
+using EntityFrameworkCore.DynamoDb.SpecificationTests.TestUtilities;
+using Microsoft.EntityFrameworkCore;
+using Microsoft.EntityFrameworkCore.Diagnostics;
+using Microsoft.EntityFrameworkCore.Metadata;
+using Microsoft.EntityFrameworkCore.Metadata.Builders;
+using Microsoft.EntityFrameworkCore.TestUtilities;
+
+namespace EntityFrameworkCore.DynamoDb.SpecificationTests;
+
+public class CustomConvertersDynamoTest(
```
ncipollina: 🚨 **Class must be `abstract` — concrete class breaks container lifetime**

Every spec test class in this project follows the pattern:
```csharp
public abstract class XxxDynamoTest : XxxTestBase<XxxDynamoTest.XxxDynamoFixture> { ... }
```
with a concrete sealed inner class carrying `[Collection(DynamoSpecificationCollection.Name)]` that injects `DynamoSpecificationContainerFixture` to anchor the shared DynamoDB Local container.

As written, `CustomConvertersDynamoTest` is a concrete class. xUnit will discover and instantiate it directly, bypassing the `DynamoSpecificationContainerFixture` entirely — the container may never start, or a new container is spun up per test class.

Fix:
1. Add `abstract` to the class declaration.
2. Add a sealed inner concrete class:
```csharp
[Collection(DynamoSpecificationCollection.Name)]
public sealed class CustomConvertersDynamoTestDefault : CustomConvertersDynamoTest
{
    public CustomConvertersDynamoTestDefault(
        CustomConvertersDynamoFixture fixture,
        DynamoSpecificationContainerFixture containerFixture)
        : base(fixture)
        => _ = containerFixture;
}
```
See `NorthwindChangeTrackingQueryDynamoTest` for the canonical pattern.

tests/EntityFrameworkCore.DynamoDb.SpecificationTests/CustomConvertersDynamoTest.cs
```diff
@@ -0,0 +1,710 @@
...
+                .Where(e => e.Id == "Gumball!!")
+                .ToListAsync()).Single();
+
+            Assert.Equal("Gumball!!", entity.Id);
+        }
+    }
+
+    public override async Task Can_insert_and_read_back_with_string_list()
+    {
+        using (var context = CreateContext())
```
ncipollina: 💧 **`using` should be `await using` — sync disposal on async context**

This block and the one at line 364 use `using (var context = CreateContext())` while every other multi-context block in this file correctly uses `await using`. The `DbContext` lifecycle in this provider supports `DisposeAsync`; sync disposal here is inconsistent and may suppress async cleanup.

Change both to `await using`.

src/EntityFrameworkCore.DynamoDb/Query/Internal/DynamoQueryableMethodTranslatingExpressionVisitor.cs
```diff
@@ -906,8 +907,9 @@ protected override ShapedQueryExpression TranslateSelect(
         if (translation != null)
             return translation;
 
-        AddTranslationErrorDetails(
-            _sqlTranslator.TranslationErrorDetails ?? DynamoStrings.PredicateNotTranslatable);
+        if (_sqlTranslator.TranslationErrorDetails is { } translationErrorDetails)
```
ncipollina: 💡 **Suggestion: restore the fallback error detail**

Before this change, when `_sqlTranslator.TranslationErrorDetails` was null, `AddTranslationErrorDetails` was still called with `DynamoStrings.PredicateNotTranslatable` as a fallback, so the final `InvalidOperationException` always included at least one context string. Now a null `TranslationErrorDetails` means no detail is added and the EF Core `CoreStrings.TranslationFailed(...)` wrapper is the only diagnostic.

Consider restoring the fallback:
```csharp
AddTranslationErrorDetails(
    _sqlTranslator.TranslationErrorDetails ?? DynamoStrings.PredicateNotTranslatable);
```

tests/EntityFrameworkCore.DynamoDb.Tests/Query/PrimitiveCollectionContainsTranslationTests.cs
```diff
@@ -0,0 +1,168 @@
...
+        var act = () => context
+            .Items
+            .AllowScan()
+            .Where(e => e.CsvStatuses.Contains(status))
+            .ToListAsync(TestContext.Current.CancellationToken);
+
+        await act
+            .Should()
+            .ThrowAsync<InvalidOperationException>()
+            .WithMessage("*Contains translation supports in-memory collection membership only*");
```
ncipollina: 💡 **Suggestion: pin assertion to the source constant**

The `WithMessage` wildcard string is a hard-coded copy of the value in `DynamoStrings.ContainsCollectionShapeNotSupported`. If that constant is ever renamed or the message changes, this test silently passes on a different message.

Couple the assertion to the source:
```csharp
.WithMessage($"*{DynamoStrings.ContainsCollectionShapeNotSupported}*");
```
This keeps the assertion in sync with the string it is testing.

tests/EntityFrameworkCore.DynamoDb.Tests/Query/QueryTypeMappingInferenceTests.cs
```diff
@@ -357,6 +475,196 @@ await context
...
+        await using var context = QueryTypeMappingContext.Create(client);
+        var code = new QueryTypeMappingCode("A-1");
+
+        var act = () => context
+            .Items
+            .AllowScan()
+            .Where(e => e.Code.Equals(code))
+            .ToListAsync(TestContext.Current.CancellationToken);
+
+        await act.Should().ThrowAsync<InvalidOperationException>();
```
ncipollina: 🔎 **Nit: assertion does not pin the failure reason**

`await act.Should().ThrowAsync<InvalidOperationException>()` with no `WithMessage` will pass for *any* `InvalidOperationException` thrown anywhere in the EF pipeline, not just a translation failure. Add a message assertion to ensure the right code path failed, consistent with the pattern used in the surrounding tests:
```csharp
await act.Should().ThrowAsync<InvalidOperationException>()
    .WithMessage("*<expected substring>*");
```

tests/EntityFrameworkCore.DynamoDb.SpecificationTests/CustomConvertersDynamoTest.cs
j-d-ha: Fixed. `CustomConvertersDynamoTest` is now abstract, with a sealed `CustomConvertersDynamoTestDefault` inner class on `DynamoSpecificationCollection` that anchors `DynamoSpecificationContainerFixture`.

tests/EntityFrameworkCore.DynamoDb.SpecificationTests/CustomConvertersDynamoTest.cs
j-d-ha: Fixed in 8cffa5e — both contexts now use `await using`.

src/EntityFrameworkCore.DynamoDb/Query/Internal/DynamoQueryableMethodTranslatingExpressionVisitor.cs
j-d-ha: This was intentional to keep the EF Core specification test behavior.

Restoring the fallback changes failures with no SQL-translator-specific detail from the plain `CoreStrings.TranslationFailed(...)` message to the "with details" shape:

```text
Additional information: The predicate could not be translated to DynamoDB PartiQL.
```

That breaks `CustomConvertersDynamoTest.Collection_property_as_scalar_Count_member`, whose base spec test asserts the exact default EF Core translation failure message. Since `_sqlTranslator.TranslationErrorDetails` is null in that path, adding generic `PredicateNotTranslatable` would make the provider produce extra diagnostic text where EF Core expects none. So this only forwards details when the SQL translator produced specific details.

src/EntityFrameworkCore.DynamoDb/Query/Internal/DynamoQueryableMethodTranslatingExpressionVisitor.cs
```diff
@@ -906,8 +907,9 @@ protected override ShapedQueryExpression TranslateSelect(
...
+            AddTranslationErrorDetails(translationErrorDetails);
```
ncipollina: 🐛 **Regression: fallback error detail removed**

The fix introduced a regression. When `_sqlTranslator.TranslationErrorDetails` is `null` — which happens for many untranslatable predicates — `AddTranslationErrorDetails` is now never called. Callers receive a generic EF Core "could not be translated" message with zero provider context.

The original code always provided a fallback:
```csharp
// Before (correct):
AddTranslationErrorDetails(
    _sqlTranslator.TranslationErrorDetails ?? DynamoStrings.PredicateNotTranslatable);
```

The `if`-guard approach should be reverted to restore the fallback:
```csharp
AddTranslationErrorDetails(
    _sqlTranslator.TranslationErrorDetails ?? DynamoStrings.PredicateNotTranslatable);
```

tests/EntityFrameworkCore.DynamoDb.Tests/Query/QueryTypeMappingInferenceTests.cs
```diff
@@ -357,6 +475,199 @@ await context
...
+        var act = () => context
+            .Items
+            .AllowScan()
+            .Where(e => e.Code.Equals(code))
+            .ToListAsync(TestContext.Current.CancellationToken);
+
+        await act
+            .Should()
+            .ThrowAsync<InvalidOperationException>()
+            .WithMessage("*Method calls are not supported*predicate translation*");
```
ncipollina: 🔎 **Nit: pin `WithMessage` to the source constant**

The glob string is a copy of `DynamoStrings.MethodCallInPredicateNotSupported`. If that constant is ever rephrased, the test silently passes on a different message.

Pin the assertion to the source, consistent with how `PrimitiveCollectionContainsTranslationTests` was fixed:
```csharp
.WithMessage($"*{DynamoStrings.MethodCallInPredicateNotSupported}*");
```

ncipollina (changes requested):
🔁 **Re-review: 1 regression, 1 nit**

Good progress — all prior blocking issues (stack imbalance, type guard, abstract class, await using) are resolved. One regression was introduced and one nit remains before this is merge-ready.

ncipollina (approved):
✅ **Approved** — both fixes confirmed correct. Build clean, all tests pass. Nice work on the full contribution.

PR #250, opened 2026-05-24
feat(query): add Single support

j-d-ha:
## Summary
Adds translation and execution support for key-only `Single` queries, including continuation token guarding for unique-result semantics. Also treats partition-key `IN` predicates as keyed access so safe query paths are not rejected as scans.

## Changes
- Translate key-only `Single` queries through DynamoDB query paths.
- Guard `Single` continuation tokens to detect multiple matching pages.
- Treat partition-key `IN` as keyed access in scan safety checks.
- Add unit/integration coverage for `Single` semantics and partition-key `IN` safety.
- Update query, diagnostics, limitations, and related docs.

## Validation
- `dotnet test --project tests/EntityFrameworkCore.DynamoDb.Tests/EntityFrameworkCore.DynamoDb.Tests.csproj --no-restore -- --output Normal`
  - Passed: 681 succeeded, 2 skipped.

## Release Notes
- `Single` queries over key predicates are now supported.
- Partition-key `IN` filters are recognized as keyed access instead of unsafe scans.


ncipollina (approved):
☑️ LGTM

PR #252, opened 2026-05-24
test(dynamodb): add value converter spec coverage

j-d-ha:
## 📋 Summary

Adds DynamoDB coverage for EF Core's end-to-end value converter specification test. The new spec test exposed missing null-converter handling in the provider, so this PR includes the small runtime fix needed for the coverage to pass.

## 📝 Changes

- Add `ValueConvertersEndToEndDynamoTest` for insert/readback value converter coverage.
- Register `ValueConvertersEndToEndTestBase<>` in `ComplianceDynamoTest`.
- Track `ValueConvertersEndToEndTestBase` as implemented in `docs/spec-test-coverage.md`.
- Document spec-test tracking expectations in `tests/EntityFrameworkCore.DynamoDb.SpecificationTests/AGENTS.md`.
- Support DynamoDB `NULL` attributes with configured converters that convert nulls.
- Preserve nullable projection behavior when removing DynamoDB projection bindings.

## 🧪 Validation

- `dotnet test --project tests/EntityFrameworkCore.DynamoDb.SpecificationTests/EntityFrameworkCore.DynamoDb.SpecificationTests.csproj --filter "FullyQualifiedName~ComplianceDynamoTest"`

### DIFF
--- a/tests/EntityFrameworkCore.DynamoDb.SpecificationTests/ValueConvertersEndToEndDynamoTest.cs
+++ b/tests/EntityFrameworkCore.DynamoDb.SpecificationTests/ValueConvertersEndToEndDynamoTest.cs
@@ -1,0 +1,50 @@
+using EntityFrameworkCore.DynamoDb.SpecificationTests.TestUtilities;
+using Microsoft.EntityFrameworkCore;
+using Microsoft.EntityFrameworkCore.TestUtilities;
+
+namespace EntityFrameworkCore.DynamoDb.SpecificationTests;
+
+public abstract class ValueConvertersEndToEndDynamoTest(
+    ValueConvertersEndToEndDynamoTest.ValueConvertersEndToEndDynamoFixture fixture)
+    : ValueConvertersEndToEndTestBase<
+        ValueConvertersEndToEndDynamoTest.ValueConvertersEndToEndDynamoFixture>(fixture)
+{
+    public override Task Can_insert_and_read_back_with_conversions(int[] valueOrder)
+        => base.Can_insert_and_read_back_with_conversions(valueOrder);
+
+    public class ValueConvertersEndToEndDynamoFixture
+        : ValueConvertersEndToEndFixtureBase, IDynamoSpecificationFixture
+    {
+        public TestSqlLoggerFactory TestSqlLoggerFactory => (TestSqlLoggerFactory)ListLoggerFactory;
+
+        protected override ITestStoreFactory TestStoreFactory => DynamoTestStoreFactory.Instance;
+
+        protected override bool ShouldLogCategory(string logCategory)
+            => DynamoSpecificationFixtureExtensions.ShouldLogDynamoSql(logCategory);
+
+        public override DbContextOptionsBuilder AddOptions(DbContextOptionsBuilder builder)
+            => base
+                .AddOptions(builder)
+                .UseDynamo(options
+                    => options.DynamoDbClient(DynamoTestStoreFactory.Instance.Client));
+
+        protected override void OnModelCreating(ModelBuilder modelBuilder, DbContext context)
+        {
+            base.OnModelCreating(modelBuilder, context);
+
+            modelBuilder
+                .Entity<ConvertingEntity>()
+                .ToTable("ConvertingEntities")
+                .HasPartitionKey(e => e.Id);
+        }
+    }
+
+    [Collection(DynamoSpecificationCollection.Name)]
+    public sealed class ValueConvertersEndToEndDynamoTestDefault : ValueConvertersEndToEndDynamoTest
+    {
+        public ValueConvertersEndToEndDynamoTestDefault(
+            ValueConvertersEndToEndDynamoFixture fixture,
+            DynamoSpecificationContainerFixture containerFixture) : base(fixture)
+            => _ = containerFixture;
+    }
+}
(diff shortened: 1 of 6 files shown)


tests/EntityFrameworkCore.DynamoDb.SpecificationTests/ValueConvertersEndToEndDynamoTest.cs
```diff
@@ -0,0 +1,50 @@
+using EntityFrameworkCore.DynamoDb.SpecificationTests.TestUtilities;
+using Microsoft.EntityFrameworkCore;
+using Microsoft.EntityFrameworkCore.TestUtilities;
+
+namespace EntityFrameworkCore.DynamoDb.SpecificationTests;
+
+public abstract class ValueConvertersEndToEndDynamoTest(
+    ValueConvertersEndToEndDynamoTest.ValueConvertersEndToEndDynamoFixture fixture)
```
ncipollina, line 8: 🐛 **Missing `Check_all_tests_overridden` and `fixture.ClearSql()` in constructor**

Every spec test class requires this sentinel method so that new methods added to the base class in future EF Core releases are caught automatically:

```csharp
[ConditionalFact]
public virtual void Check_all_tests_overridden()
    => DynamoTestHelpers.AssertAllTestMethodsOverridden(typeof(ValueConvertersEndToEndDynamoTest));
```

The constructor also needs to call `fixture.ClearSql()` to flush any SQL captured during fixture initialization before `AssertSql` assertions run. See `FindDynamoTest`, `CustomConvertersDynamoTest`, etc. for the pattern:

```csharp
protected ValueConvertersEndToEndDynamoTest(ValueConvertersEndToEndDynamoFixture fixture)
    : base(fixture)
    => fixture.ClearSql();
```

tests/EntityFrameworkCore.DynamoDb.SpecificationTests/ValueConvertersEndToEndDynamoTest.cs
```diff
@@ -0,0 +1,50 @@
...
+    : ValueConvertersEndToEndTestBase<
+        ValueConvertersEndToEndDynamoTest.ValueConvertersEndToEndDynamoFixture>(fixture)
+{
+    public override Task Can_insert_and_read_back_with_conversions(int[] valueOrder)
+        => base.Can_insert_and_read_back_with_conversions(valueOrder);
+
```
ncipollina, line 14: 💡 **Consider adding `AssertSql` or a comment explaining deferral**

This test passes (it was the driver for the runtime fix), so it should either assert the PartiQL it emits or explain why SQL baselining is deferred:

```csharp
public override async Task Can_insert_and_read_back_with_conversions(int[] valueOrder)
{
    await base.Can_insert_and_read_back_with_conversions(valueOrder);
    AssertSql(/* ... */);
}
```

If the statement count is large and baselining is intentionally deferred, a brief comment to that effect is enough.

tests/EntityFrameworkCore.DynamoDb.SpecificationTests/ValueConvertersEndToEndDynamoTest.cs
```diff
@@ -0,0 +1,50 @@
...
+        {
+            base.OnModelCreating(modelBuilder, context);
+
+            modelBuilder
+                .Entity<ConvertingEntity>()
+                .ToTable("ConvertingEntities")
+                .HasPartitionKey(e => e.Id);
+        }
+    }
+
```
ncipollina, line 41: 🐛 **`CleanAsync` not overridden — will cause flaky failures on repeated theory runs**

`Can_insert_and_read_back_with_conversions` is a `[ConditionalTheory]` parameterized with `int[] valueOrder`, so xUnit runs it multiple times. Without this override, tables are not reset between runs and entity ID collisions from a previous run will cause failures.

Every other DynamoDB fixture that mutates data overrides this — add it inside `ValueConvertersEndToEndDynamoFixture`:

```csharp
protected override async Task CleanAsync(DbContext context)
{
    await context.Database.EnsureDeletedAsync();
    await context.Database.EnsureCreatedAsync();
}
```

ncipollina (changes requested):
Thanks for the contribution! The runtime fix ( + ) is correct and well-reasoned. A few things in the spec test class need to be addressed before merging.

tests/EntityFrameworkCore.DynamoDb.SpecificationTests/ValueConvertersEndToEndDynamoTest.cs
j-d-ha: This test has a bunch of non-virtual methods that we can't override, so that's why this class does not have that method.

tests/EntityFrameworkCore.DynamoDb.SpecificationTests/ValueConvertersEndToEndDynamoTest.cs
j-d-ha: Given that we can't override the methods that actually run the commands, we can't assert the generated PartiQL.

tests/EntityFrameworkCore.DynamoDb.SpecificationTests/ValueConvertersEndToEndDynamoTest.cs
j-d-ha: I don’t think this is a real issue.

`DynamoTestStore.CleanAsync` already clears the tracker and runs `EnsureDeletedAsync()` + `EnsureCreatedAsync()`. `SharedStoreFixtureBase.InitializeAsync` passes fixture `CleanAsync` into `TestStore.InitializeAsync`, and `TestStore.InitializeAsync` then calls both the fixture clean hook and the test-store clean hook during fixture initialization.

Adding this override to `ValueConvertersEndToEndDynamoFixture` would mostly duplicate the existing Dynamo test-store cleanup, and it would not run between `[ConditionalTheory]` rows anyway.

Also, `ConvertingEntity.Id` is a generated `Guid`; each theory row inserts a new entity and then queries by that generated id, so prior rows should not cause ID collisions. I ran the filtered test locally and all 3 theory rows passed.

ncipollina (approved):
✅ LGTM!!!

PR #254, opened 2026-06-07
feat: support EF key semantics for DynamoDB table keys

j-d-ha:
## Summary

Implements ADR-003 so EF key configuration can participate in DynamoDB table-key mapping. Provider APIs (`HasPartitionKey`/`HasSortKey`) remain preferred and supported, while `HasKey`, `[Key]`, and `[PrimaryKey]` now resolve to finalized partition/sort roles for validation, runtime metadata, queries, `Find`, writes, and table definition generation.

## Changes

- Added finalized DynamoDB table-key resolver and model-finalizing convention.
- Enabled one- and two-part EF keys as DynamoDB partition/sort keys, including data annotations.
- Added conflict validation for mismatched EF/provider key configuration and invalid 3+ key shapes.
- Allowed typed mapped shadow table keys while keeping runtime-only provider metadata rejected.
- Updated query planning, `Find`, write SQL, table-definition, value-generation coverage, and docs for EF-key models.
- Removed obsolete split key conventions and updated ADR/docs to match implemented behavior.

## Validation

- `dotnet test --project tests/EntityFrameworkCore.DynamoDb.Tests/EntityFrameworkCore.DynamoDb.Tests.csproj` — passed: 720 succeeded, 2 skipped.
- `uv run zensical build` — passed.
- Targeted metadata/query/storage/find/table-definition test filters were run throughout bead completion.
- Integration smoke command was attempted, but Docker/Testcontainers was unavailable in this environment (`DockerUnavailableException`).

## Release Notes

- EF-native key configuration (`HasKey`, `[Key]`, `[PrimaryKey]`) now works for DynamoDB table keys when using one or two key properties.
- Combined EF/provider key configuration must agree on partition/sort order.
- Typed mapped shadow table keys are supported; runtime-only provider metadata remains invalid as a table key.

## Notes for Reviewers

Focus review on finalized key role flow across metadata, validation, query safety, `Find`, and write/table-definition consumers. The branch also includes planning artifacts under `issue-253/` and `plans/` because they were present as user-authored worktree files when preparing the PR.


src/EntityFrameworkCore.DynamoDb/Metadata/Internal/DynamoTableKeyResolver.cs
```diff
@@ -0,0 +1,213 @@
...
+                primaryKeySource,
+                annotatedPartitionKey,
+                annotatedSortKey,
+                partitionKeySource,
+                sortKeySource);
+
+        var partitionKey = annotatedPartitionKey ?? ResolvePartitionKeyByConvention(entityType);
+        var sortKey = annotatedSortKey ?? ResolveSortKeyByConvention(entityType);
+
+        if (partitionKey is null
```
ncipollina, line 46: ✨ The three-tier resolution precedence in `Resolve()` is clean and well-reasoned:

1. Explicit EF primary key (`HasKey`/`[Key]`/`[PrimaryKey]`) → `ResolveExplicitPrimaryKey`
2. Provider annotations (`HasPartitionKey`/`HasSortKey`) → conventional DynamoDB name fallback (`PK`/`PartitionKey`/`Id`, `SK`/`SortKey`)
3. Convention-level EF PK (only when partition key is still unresolved after steps 1 and 2)

Regarding the question in the PR about whether DynamoDB conventional names win over EF convention PK — yes, they do. The `partitionKey is null` guard on line 46 means the convention-PK fallback block only activates when `ResolvePartitionKeyByConvention` returned nothing (no DynamoDB-named property found). This is the correct ordering.

src/EntityFrameworkCore.DynamoDb/Metadata/Internal/DynamoTableKeyResolver.cs
```diff
@@ -0,0 +1,213 @@
...
+        if (primaryKey.Properties.Count > 2)
+            throw new InvalidOperationException(
+                $"Entity type '{entityType.DisplayName()}' has EF primary key "
+                + $"[{FormatProperties(primaryKey.Properties)}], but DynamoDB table keys support only "
+                + "one- or two-part keys: partition key with optional sort key.");
+
+        var primaryPartitionKey = primaryKey.Properties[0];
+        var primarySortKey = primaryKey.Properties.Count == 2 ? primaryKey.Properties[1] : null;
+
+        if (annotatedPartitionKey is not null && annotatedPartitionKey == annotatedSortKey)
```
ncipollina, line 84: ❓ The early `ValidateResolvedRoles` call here is only guarding the case where the same property is simultaneously annotated as both partition key and sort key (`annotatedPartitionKey == annotatedSortKey`). This works correctly — `ValidateResolvedRoles` throws on equal properties — but the condition reads like a guard rather than a validation trigger.

A small readability improvement would be to invert the guard into the standard always-validate-then-return pattern, or add a clarifying comment like:

```csharp
// Both annotations point to the same property — throw before the conflict checks below.
if (annotatedPartitionKey is not null && annotatedPartitionKey == annotatedSortKey)
    ValidateResolvedRoles(entityType, annotatedPartitionKey, annotatedSortKey);
```

Not a bug — the check is correct and `ValidateResolvedRoles` will throw, so execution never reaches the later conflict checks. Just noting this for future readers.

src/EntityFrameworkCore.DynamoDb/Metadata/Conventions/DynamoKeyDiscoveryConvention.cs
```diff
@@ -5,34 +5,13 @@
 
 namespace EntityFrameworkCore.DynamoDb.Metadata.Conventions;
 
-/// <summary>A convention that discovers DynamoDB table keys from CLR property naming conventions.</summary>
+/// <summary>Legacy helper for DynamoDB conventional key-name matching.</summary>
```
ncipollina: ⚠️ `DynamoKeyDiscoveryConvention` still inherits `KeyDiscoveryConvention` and overrides `ShouldDiscoverKeyProperties` and `ProcessKeyProperties`, but this class is no longer registered in the convention pipeline (the `Replace<KeyDiscoveryConvention>` call was removed). The overrides are now **dead code** — EF's base `KeyDiscoveryConvention` runs instead.

The new remarks doc acknowledges this. Worth calling out that the `ShouldDiscoverKeyProperties` override (which skips owned types) is no longer active. This is safe because `DynamoOwnedEntityTypeValidationConvention` still rejects owned types at finalization. But it would avoid future confusion to note this explicitly, or to remove the overrides if the class is purely a static helper now.

💡 Consider whether `DynamoKeyDiscoveryConvention` should simply be made a non-inheriting static class holding only the helper methods (`GetPartitionKeyCandidates`, `IsDiscoverableKeyProperty`, etc.), since the `KeyDiscoveryConvention` base and the overrides serve no purpose anymore.

src/EntityFrameworkCore.DynamoDb/Storage/DynamoEntityWritePlanFactory.cs
```diff
@@ -60,7 +59,7 @@ internal AttributeValue SerializeProperty(IUpdateEntry entry, IProperty property
             throw new InvalidOperationException(
```
ncipollina, line 59: ✨ Removing `.Where(static p => !(p.IsShadowProperty() && p.IsKey()))` is the right call now that mapped shadow keys (e.g. `b.Property<string>("PK").HasKey("PK")`) should round-trip through DynamoDB. The updated error message is accurate: the only properties now excluded from the write plan are runtime-only ones (the `__executeStatementResponse` shadow added by `DynamoResponseShadowPropertyConvention`), and `SaveChanges_AddedShadowHasKeyEntity_IncludesShadowKeyAttributes` proves it.

💡 Minor nit: the `"Runtime-only properties are not serialized"` tail in the error could misguide a caller who passes a property from a different entity type (not a bug, but the message implies the callers property is runtime-only). Consider a slightly more accurate tail: `"Only runtime-only properties are excluded from the write plan."`

tests/EntityFrameworkCore.DynamoDb.Tests/Storage/DynamoTableKeySaveChangesTests.cs
```diff
@@ -0,0 +1,431 @@
...
+        var (context, captured) = CreateContext();
+        context.Add(
+            new DerivedDocument { Pk = "DOC#1", Sk = "META#1", Name = "before", Extra = "extra" });
+
+        await context.SaveChangesAsync(TestContext.Current.CancellationToken);
+
+        AssertStatement(
+            captured,
+            """
+            INSERT INTO "Documents"
```
ncipollina, line 28: 💡 The old `DynamoDerivedSaveChangesTests` INSERT test used `.Contain()` (order-agnostic) for the `VALUE` attributes. This new test locks in the column order: `pk_attr, sk_attr, $type, name, extra`.

This is stricter and more precise — it documents that the discriminator shadow property is enumerated between `sk_attr` and `name` (before other base CLR properties). That ordering is driven by EF Core's internal property ordinal assignment, which assigns shadow properties added by `HasDiscriminator<>()` an ordinal after explicitly-mapped CLR properties of the base type… but the ordering here puts `$type` _before_ `name`.

This passes in CI, so the ordering is stable today. It's worth adding a brief comment like:

```csharp
// Column order: EF assigns discriminator shadow property ordinal before CLR properties
// of the same entity type when registered via HasDiscriminator<>().
```

…so that if a future EF Core update shifts ordinal assignment, the failure is easy to diagnose rather than a mysterious string mismatch.

src/EntityFrameworkCore.DynamoDb/Metadata/Conventions/DynamoConventionSetBuilder.cs
```diff
@@ -41,14 +39,10 @@ public override ConventionSet CreateConventionSet()
...
-        // Must run before DynamoKeyAnnotationConvention so PK/SK attribute names are already
+        // Must run before DynamoTableKeyResolutionConvention so PK/SK attribute names are already
         // transformed when key validation reads them via GetAttributeName().
         conventionSet.ModelFinalizingConventions.Add(new DynamoAttributeNamingConventionApplier());
-        conventionSet.ModelFinalizingConventions.Add(new DynamoKeyAnnotationConvention());
-
-        var keyInPrimaryKeyConvention = new DynamoKeyInPrimaryKeyConvention(Dependencies);
-        conventionSet.EntityTypeAnnotationChangedConventions.Add(keyInPrimaryKeyConvention);
-        conventionSet.PropertyAddedConventions.Add(keyInPrimaryKeyConvention);
+        conventionSet.ModelFinalizingConventions.Add(new DynamoTableKeyResolutionConvention());
```
ncipollina, line 45: ✨ Excellent consolidation. Replacing three incremental conventions (`DynamoKeyAnnotationConvention`, `DynamoKeyInPrimaryKeyConvention`, and the `Replace<KeyDiscoveryConvention>`) with a single `IModelFinalizingConvention` dramatically simplifies the key-resolution flow. The old approach relied on interleaved incremental triggers (property-added, annotation-changed) that were order-sensitive and hard to reason about at a glance. The new single-pass finalization over the complete, stable model is much easier to test exhaustively — which the 737-test pass count reflects.

src/EntityFrameworkCore.DynamoDb/Metadata/Internal/DynamoResolvedTableKey.cs
```diff
@@ -0,0 +1,12 @@
+using Microsoft.EntityFrameworkCore.Metadata;
+
+namespace EntityFrameworkCore.DynamoDb.Metadata.Internal;
+
+/// <summary>Resolved DynamoDB table-key metadata for a root entity type.</summary>
+internal sealed record DynamoResolvedTableKey(
```
ncipollina, line 14: ❓ The question raised in the PR summary — "When `PrimaryKeyConfigurationSource` is null (no primary key), `Convention.Overrides(null)` returns true so the code would try to rebuild the EF PK" — is answered by the short-circuit on line 44: `resolvedKey.PrimaryKey != null &&`. When there is no primary key, the whole block is skipped and the code falls through to `entityType.Builder.PrimaryKey(keyProperties)` unconditionally (assuming `PartitionKey` is non-null and `entityType.Builder` is non-null).

Since `PrimaryKey` and `PrimaryKeyConfigurationSource` are always paired in the DTO (both from `primaryKey?.GetConfigurationSource()`), `PrimaryKey != null` implies `PrimaryKeyConfigurationSource != null`. The null safety here is correct, but it's a subtle invariant worth a short comment on `DynamoResolvedTableKey` to state that `PrimaryKeyConfigurationSource` is non-null if-and-only-if `PrimaryKey` is non-null.

ncipollina (dismissed):
## Review: feat/add-ef-core-key — EF Key Semantics for DynamoDB Table Keys

This is a well-executed architectural simplification. The core logic is correct, the test coverage is thorough (735 passing, 2 intentionally skipped, 0 failed), and the behavioral change is non-breaking for existing users.

### What works well

- **Convention consolidation** is the standout win. Three interleaved incremental conventions (`DynamoKeyAnnotationConvention`, `DynamoKeyInPrimaryKeyConvention`, `Replace<KeyDiscoveryConvention>`) replaced by a single `IModelFinalizingConvention` that operates on a stable, complete model. This is the right pattern for complex multi-source resolution.
- **Resolution precedence** in `DynamoTableKeyResolver.Resolve()` is correct. The DynamoDB conventional names (`PK`/`PartitionKey`/`Id`, `SK`/`SortKey`) always take priority over EF's convention-level PK — the third-tier fallback only activates when both annotation and conventional-name resolution return null.
- **Shadow key serialization** — removing the `IsShadowProperty() && IsKey()` filter and relying solely on `IsRuntimeOnly()` is the right semantic. The `DynamoResponseShadowPropertyConvention` shadow property is the only shadow property that should be excluded, and it's already marked `IsRuntimeOnly`.
- **Validation removal** — dropping `ValidateRootEntityDoesNotUseExplicitPrimaryKeyConfiguration` and the shadow-key rejection from `ValidateConfiguredKeyPropertyIsSupported` unblocks the new scenarios without leaving gaps; the remaining validators still enforce partition key presence, sort key pairing, type compatibility, and nullability.
- **`DynamoResolvedTableKey` record** — compact, immutable, makes the data flow through `ApplyResolvedKey` easy to trace.

### Findings

**No blocking bugs found.** All issues below are minor:

1. **`DynamoKeyDiscoveryConvention` dead code** (⚠️ minor): The class still inherits `KeyDiscoveryConvention` and carries active overrides (`ShouldDiscoverKeyProperties`, `ProcessKeyProperties`), but nothing registers it in the convention pipeline anymore. These overrides are dead code. Consider removing the inheritance and overrides, making this a pure static helper. The `ShouldDiscoverKeyProperties` owned-type guard is safe to drop because `DynamoOwnedEntityTypeValidationConvention` still catches owned types at finalization. (Inline comment posted.)

2. **Early `ValidateResolvedRoles` guard readability** (💡 cosmetic): Lines 83–84 in `ResolveExplicitPrimaryKey` — the early call is correct (throws immediately when `annotatedPartitionKey == annotatedSortKey`) but reads ambiguously. A one-line comment would help future readers. (Inline comment posted.)

3. **`DynamoResolvedTableKey` invariant** (❓ documentation): `PrimaryKeyConfigurationSource` is nullable but is always non-null when `PrimaryKey` is non-null (they're paired from `primaryKey?.GetConfigurationSource()`). The `ApplyResolvedKey` null handling is correct due to the `PrimaryKey != null &&` short-circuit, but this invariant isn't documented. (Inline comment posted.)

4. **`SerializeProperty` error message** (💡 cosmetic): The `"Runtime-only properties are not serialized"` tail is accurate for the expected failure path but could confuse a caller passing a property from a different entity type entirely. Minor wording suggestion in inline comment.

5. **INSERT column-order assertion** (💡 minor): The new test pins the discriminator to appear before `name` in the `VALUE` map. This is stable today and the test passes, but a clarifying comment on why `$type` appears before `name` would help future EF Core version bumps. (Inline comment posted.)

### Derived entity test coverage

The old `DynamoDerivedSaveChangesTests` file is deleted and all its tests are preserved in `DynamoTableKeySaveChangesTests`. Coverage is fully intact:
- INSERT for derived entity ✓ (now with exact column order rather than `.Contain()`)
- UPDATE WHERE clause for derived entity ✓
- DELETE WHERE clause for derived entity ✓
- Shared table group annotations ✓
- New: HasKey-only, shadow key, single-property HasKey scenarios ✓

### Verdict

Approved. The five items above are all non-blocking style/documentation notes. No correctness bugs, no safety gaps, no API breaks.

src/EntityFrameworkCore.DynamoDb/Metadata/Internal/DynamoTableKeyResolver.cs
j-d-ha: Added a comment to make it clearer.

src/EntityFrameworkCore.DynamoDb/Metadata/Conventions/DynamoKeyDiscoveryConvention.cs
j-d-ha: Fixed!

src/EntityFrameworkCore.DynamoDb/Storage/DynamoEntityWritePlanFactory.cs
j-d-ha: Fixed

tests/EntityFrameworkCore.DynamoDb.Tests/Storage/DynamoTableKeySaveChangesTests.cs
j-d-ha: Fixed

src/EntityFrameworkCore.DynamoDb/Metadata/Internal/DynamoResolvedTableKey.cs
j-d-ha: Added a comment to explain this better.

ncipollina (dismissed):
All feedback addressed — re-reviewing and approving. ✅

- ** dead inheritance** — class deleted; static helpers cleanly extracted to . Exactly the right move.
- **Early  guard** — clarifying comment added; intent is now unambiguous.
- ** invariant** — XML doc documents the  non-null iff  non-null invariant.
- **Serializer error message** — updated to `"Only runtime-only properties are excluded from the write plan."` — accurate and actionable.

No new concerns. LGTM. 🚀

PR #258, opened 2026-06-18
feat(query): add scalar translation spec coverage

j-d-ha:
## Summary
Adds DynamoDB query translation coverage for GUIDs, enums, strings, arithmetic skips, and complex-property structural equality. Also expands string translation behavior for `Length`, `IsNullOrEmpty`, and compare-to-constant forms, with docs updated to match supported and unsupported query behavior.

## Changes
- Add specification tests for GUID, enum, string, and complex-property structural equality translations.
- Translate string `Length` via PartiQL `size(...)` and `string.IsNullOrEmpty(...)` via null, missing, or empty-string checks.
- Broaden string comparison translation for `Compare`/`CompareTo` against `-1`, `0`, and `1`.
- Update query docs, limitations, and spec coverage notes for supported and explicitly unsupported translations.

## Validation
- Not run during PR creation. PR opened from existing remote branch only, per request.

## Notes for Reviewers
Local working tree has uncommitted/untracked files; none were pushed or included. PR uses `origin/spec/guid-translations` as-is.


src/EntityFrameworkCore.DynamoDb/Query/Internal/DynamoSqlTranslatingExpressionVisitor.cs
```diff
@@ -1139,39 +1160,68 @@ private Expression TranslateDynamoDbFunctions(MethodCallExpression node)
...
+        var result = sqlExpressionFactory.Binary(binaryOp.Value, a, b);
         if (result is not null)
             return result;
 
-        AddTranslationErrorDetails(DynamoStrings.UnsupportedBinaryOperator(effectiveOp));
+        AddTranslationErrorDetails(DynamoStrings.UnsupportedBinaryOperator(binaryOp.Value));
         return QueryCompilationContext.NotTranslatedExpression;
     }
 
+    private static ExpressionType? TryMapStringCompareOperator(ExpressionType op, int comparand)
```
ncipollina, line 1203: **Correctness: operator mapping table is correct.**

Each arm has been verified against sign-semantics reasoning (treating 1 as "any positive" and -1 as "any negative"):

| comparand | op | effectiveOp | rationale |
|---|---|---|---|
| 1 | `==` or `>=` | `>` | result is positive → a > b |
| 1 | `!=` or `<` | `<=` | result is not positive → a ≤ b |
| 1 | `>` or `<=` | null | always-false / always-true — correctly rejected |
| -1 | `==` or `<=` | `<` | result is negative → a < b |
| -1 | `!=` or `>` | `>=` | result is not negative → a ≥ b |
| -1 | `<` or `>=` | null | always-false / always-true — correctly rejected |

Also confirmed: out-of-range comparands (e.g., 42) fall through the outer `_ => null` arm, causing `TryTranslateStringCompare` to return null and correctly delegating to client-side evaluation. The `Compare_simple_more_than_one` skip reason is accurate.

src/EntityFrameworkCore.DynamoDb/Query/Internal/DynamoSqlTranslatingExpressionVisitor.cs
```diff
@@ -1209,9 +1259,27 @@ private static ExpressionType FlipComparison(ExpressionType op)
             ExpressionType.GreaterThanOrEqual => ExpressionType.LessThanOrEqual,
             ExpressionType.LessThan => ExpressionType.GreaterThan,
             ExpressionType.LessThanOrEqual => ExpressionType.GreaterThanOrEqual,
-            _ => op,
+            _ => op
         };
 
+    /// <summary>Translates string.IsNullOrEmpty to null, missing, or empty-string checks.</summary>
+    private Expression TranslateStringIsNullOrEmpty(MethodCallExpression node)
```
ncipollina, line 1266: **Expression reuse is safe here.**

`argument` is translated once, then referenced three times across `IsNull`, `IsMissing`, and the equality `Binary`. `SqlExpression` instances in EF Core are immutable — none of `IsNull`, `IsMissing`, or `Binary` mutate the expression; they each produce a new wrapper node. Reusing the same instance across multiple tree nodes is the standard EF Core pattern (see the existing `TranslateStringContains` for the same idiom). `SqlParameterExpression` and `SqlPropertyExpression` are equally safe to share.

src/EntityFrameworkCore.DynamoDb/Query/Internal/DynamoSqlTranslatingExpressionVisitor.cs
```diff
@@ -396,6 +405,15 @@ protected override Expression VisitParameter(ParameterExpression node)
     /// <inheritdoc />
     protected override Expression VisitMember(MemberExpression node)
     {
+        if (node is { Member.Name: nameof(string.Length), Expression.Type: var expressionType }
```
ncipollina, line 408: **Nullable `string?` check is safe; minor note on `node.Expression` null dereference.**

`string?` and `string` share the same CLR type at runtime (`typeof(string?) == typeof(string)`), so the `expressionType == typeof(string)` check correctly handles nullable string properties.

One theoretical concern: if `node.Expression` is null (static member access), the nested pattern `{ Expression.Type: var expressionType }` would throw `NullReferenceException` at the property access. This is unreachable in practice — EF Core only produces `MemberExpression` nodes for instance members (`string.Length` is always instance-bound), and a null `Expression` only arises for static fields/methods, which the EF pipeline never generates for property access. No change needed, but it is worth a comment if you encounter similar patterns elsewhere in the visitor.

src/EntityFrameworkCore.DynamoDb/Query/Internal/DynamoProjectionBindingExpressionVisitor.cs
```diff
@@ -400,7 +400,8 @@ protected override Expression VisitMember(MemberExpression node)
             // nested paths are WHERE-only. Return NotTranslated so EF Core falls back to
             // client-side evaluation via the index-based binding pass.
             var translation = sqlTranslator.Translate(node);
-            if (translation is null or DynamoScalarAccessExpression)
+            if (translation is not SqlPropertyExpression
+                and not DynamoComplexPropertyAccessExpression)
```
ncipollina, line 404: **Narrowing guard is correct and intentional; one subtlety worth documenting.**

Old guard:
```csharp
if (translation is null or DynamoScalarAccessExpression)
```
New guard:
```csharp
if (translation is not SqlPropertyExpression
    and not DynamoComplexPropertyAccessExpression)
```

The old guard inadvertently allowed `SqlFunctionExpression` (e.g., `size(attr)` from `string.Length`), `SqlBinaryExpression`, and `SqlConstantExpression` through to the projection mapping. The new guard is correctly narrower: only attribute-path expressions can be projected server-side in PartiQL; computed/function expressions fall back to client evaluation via the index-based binding pass.

The change is semantically correct. Consider adding a brief comment to `VisitMember` explaining *why* the guard specifically allows only `SqlPropertyExpression | DynamoComplexPropertyAccessExpression` (i.e., PartiQL `SELECT` supports only attribute paths, not computed expressions), since the old guard's surface-level shape was misleading.

src/EntityFrameworkCore.DynamoDb/Query/Internal/DynamoSqlTranslatingExpressionVisitor.cs
```diff
@@ -43,6 +43,9 @@ public sealed class DynamoSqlTranslatingExpressionVisitor(
     private static readonly MethodInfo StringStartsWithMethod =
         ((Func<string, bool>)string.Empty.StartsWith).Method;
 
+    private static readonly MethodInfo StringIsNullOrEmptyMethod =
```
ncipollina, line 46: The null-forgiving `!` is justified — `string.IsNullOrEmpty(string)` is a well-known BCL method that definitely exists. This is consistent with the style of all other `MethodInfo` static fields in this class (`IsNullMethod`, `IsMissingMethod`, etc.). No issue here.

docs/limitations.md
```diff
@@ -217,12 +217,20 @@ whose projection type is not `ALL`. `KEYS_ONLY` and `INCLUDE` index candidates a
...
+### String Function Limitations
 
 `string.StartsWith(s)` and `string.Contains(s)` translate to `begins_with` and `contains` in
 PartiQL only for the single-`string`-argument overloads. Overloads that accept a `char`, a
 `StringComparison`, or a `CultureInfo` argument throw at translation time.
 
+DynamoDB PartiQL does not provide `ends_with`, regex, substring, replace, trim, or case-conversion
+functions, so `EndsWith`, `Regex`, `IndexOf`, `Substring`, `Replace`, `Trim*`, `ToUpper`, and
+`ToLower` are not translated.
+
```
ncipollina, line 229: Good call documenting the UTF-16 vs. DynamoDB `size()` semantics mismatch. DynamoDB's `size` for strings returns the number of UTF-8 bytes, not the .NET UTF-16 code units returned by `string.Length`. For ASCII strings they agree; for multi-byte code points (including emoji and non-BMP characters) they diverge. The warning is accurate and important for anyone relying on this for validation logic.

tests/EntityFrameworkCore.DynamoDb.SpecificationTests/Query/Translations/StringTranslationsDynamoTest.cs
```diff
@@ -0,0 +1,758 @@
...
+
+/// <summary>String translation specification tests for the DynamoDB provider.</summary>
+public abstract class StringTranslationsDynamoTest
+    : StringTranslationsTestBase<BasicTypesQueryDynamoFixture>
+{
+    protected StringTranslationsDynamoTest(BasicTypesQueryDynamoFixture fixture) : base(fixture)
+        => fixture.ClearSql();
+
+    [ConditionalFact]
+    public virtual void Check_all_tests_overridden()
```
ncipollina, line 14: The `Check_all_tests_overridden` guard is a great discipline. It ensures that newly added base test methods in future EF Core spec updates are immediately surfaced as failures rather than silently passing without a DynamoDB-specific SQL assertion or an explicit skip. Good pattern consistently applied across all three new spec files.

tests/EntityFrameworkCore.DynamoDb.SpecificationTests/Query/Translations/StringTranslationsDynamoTest.cs
```diff
@@ -0,0 +1,758 @@
...
+    public override Task Substring_with_two_args_with_parameter()
+        => base.Substring_with_two_args_with_parameter();
+
+    [ConditionalFact(
+        Skip = "DynamoDB PartiQL has no corresponding string function for this operation.")]
+    public override Task Substring_with_two_args_with_IndexOf()
+        => base.Substring_with_two_args_with_IndexOf();
+
+    public override async Task IsNullOrEmpty()
+    {
```
ncipollina, line 180: The generated PartiQL for `IsNullOrEmpty` is correct:

```sql
WHERE "string" IS NULL OR "string" IS MISSING OR "string" = ''
```

The left-to-right evaluation order (NULL → MISSING → empty) is semantically sound. DynamoDB PartiQL short-circuits `OR`, so the `IS NULL` and `IS MISSING` checks guard against attribute-absent cases before the equality comparison, which would be undefined if the attribute were missing. The double-negation in `IsNullOrEmpty_negated` correctly wraps the whole expression in `NOT (...)`.

tests/EntityFrameworkCore.DynamoDb.SpecificationTests/SkipReason.cs
```diff
@@ -12,6 +12,9 @@ public static class SkipReason
     public const string ComplexTypeSubqueriesNotSupported =
         "DynamoDB provider does not support complex type values in subqueries or Contains predicates.";
 
+    public const string ComplexCollectionStructuralEqualityNotSupported =
+        "DynamoDB provider does not support structural equality comparisons for complex collection properties. See issue #257.";
```
ncipollina, line 16: Good addition of `ComplexCollectionStructuralEqualityNotSupported` with the issue reference (`#257`). Centralizing skip reasons as `const string` values avoids drift between test files and makes it easy to grep for all tests that share a limitation. The pattern is consistent with the rest of the class.

ncipollina (approved):
## Review: feat(query): add scalar translation spec coverage

**Decision: APPROVE** — no blocking issues found. All new logic is correct.

### What was reviewed

- `TryMapStringCompareOperator` sign-to-operator mapping table (all 12 arms verified manually)
- `TranslateStringIsNullOrEmpty` expression reuse across three SQL nodes
- `VisitMember` string.Length guard for nullable strings and null `Expression`
- Projection binding guard change (`is null or DynamoScalarAccessExpression` → `is not SqlPropertyExpression and not DynamoComplexPropertyAccessExpression`)
- `StringIsNullOrEmptyMethod` static field null-forgiving operator
- All six new SQL baselines in `Compare_simple_one` and `CompareTo_simple_one` cross-verified against the EF Core base test predicates
- Docs: `limitations.md` size-semantics warning, `SkipReason.cs` additions

### Findings

**No correctness bugs.** All logic is semantically accurate. Details in inline comments:

- **`TryMapStringCompareOperator`** — operator table is correct. Sign-semantics interpretation (1 = "any positive", -1 = "any negative") matches EF Core convention. The two "always-true" arms (`> 1` and `>= -1`) correctly return null to reject translation. Out-of-range comparands (e.g. 42) fall through to the outer `_ => null` and correctly produce client-side evaluation.

- **Expression reuse in `TranslateStringIsNullOrEmpty`** — safe. `SqlExpression` instances are immutable; sharing the same reference across `IsNull`, `IsMissing`, and the `Binary` equality node is correct and consistent with existing patterns.

- **`string.Length` pattern match** — the `{ Expression.Type: var expressionType }` dereference is safe in all EF Core-generated trees (instance member → non-null `Expression`). `string?` resolves to `typeof(string)` at runtime; no issue.

- **Projection binding guard narrowing** — the new `is not SqlPropertyExpression and not DynamoComplexPropertyAccessExpression` guard is correctly tighter than the old `is null or DynamoScalarAccessExpression`. The old guard would have let `SqlFunctionExpression` (e.g., `size()`) through to projection mapping, which is wrong for PartiQL. The new guard correctly routes all non-path expressions to client-side evaluation.

### Suggestions (non-blocking)

1. Consider adding a one-line comment above the new narrowing guard in `DynamoProjectionBindingExpressionVisitor.VisitMember` explaining that PartiQL `SELECT` only supports attribute paths — the old guard's shape was misleading and a new reader may find the explicit allow-list surprising without context.

2. The limitation doc correctly notes the UTF-16 vs. UTF-8 byte semantics mismatch for `string.Length → size()`. No action needed, but worth verifying against DynamoDB's own documentation that `size` for string items returns UTF-8 encoded byte length (this is consistent with AWS docs).

### What is well done

- Comprehensive `AssertSql` baseline coverage for 758 lines of new spec tests
- `Check_all_tests_overridden` guard applied consistently across all three new test classes
- Centralized `SkipReason` constants with issue reference (`#257`)
- Skip messages are precise about the specific limitation (overloads vs. missing function vs. sign semantics)
- Trailing-comma cleanup throughout switch expressions is clean and consistent

PR #259, opened 2026-06-18
fix(query): translate nullable complex value members

j-d-ha:
## Summary

Adds translation support and coverage for nullable complex value-object member access. Nullable complex `HasValue` now translates to DynamoDB null/missing existence checks, and `.Value.Property` resolves to the expected nested attribute path.

## Changes

- Reuses normalized member-chain names for direct property access so nullable `.Value` wrappers do not leak into model property lookup.
- Adds specification coverage and `AssertSql` baselines for nullable complex `.Value.Property` and `HasValue` queries.
- Documents nullable complex value-object filtering and supported operator behavior.

## Validation

- `dotnet test tests/EntityFrameworkCore.DynamoDb.SpecificationTests/EntityFrameworkCore.DynamoDb.SpecificationTests.csproj --filter 'FullyQualifiedName~ComplexPropertiesMiscellaneousDynamoTest' --no-restore`
- `task docs:build`
- `git diff --check`

## Release Notes

- Query translation now supports nullable complex value-object `.Value.Property` access and `HasValue` checks.


src/EntityFrameworkCore.DynamoDb/Query/Internal/DynamoSqlTranslatingExpressionVisitor.cs
```diff
@@ -367,6 +375,11 @@ private bool TryGetMemberAccessChain(
         return sqlExpression;
     }
 
+    private static bool IsNullableMember(MemberInfo member, string memberName)
+        => member.Name == memberName
+            && member.DeclaringType is { IsGenericType: true } declaringType
+            && declaringType.GetGenericTypeDefinition() == typeof(Nullable<>);
```
ncipollina, line 384: Well-scoped guard. The `GetGenericTypeDefinition() == typeof(Nullable<>)` check is the correct defense against user-defined types that happen to declare a `Value` or `HasValue` member. No false positives possible here.

src/EntityFrameworkCore.DynamoDb/Query/Internal/DynamoSqlTranslatingExpressionVisitor.cs
```diff
@@ -396,36 +409,29 @@ protected override Expression VisitParameter(ParameterExpression node)
...
-                break;
-            }
+        if (IsNullableMember(node.Member, nameof(Nullable<int>.HasValue))
+            && node.Expression is { } nullableExpression
+            && TryTranslateComplexPropertyForStructuralComparison(nullableExpression) is
+                { } complexOperand)
+            return sqlExpressionFactory.Binary(
+                ExpressionType.AndAlso,
+                sqlExpressionFactory.IsNotNull(complexOperand),
+                sqlExpressionFactory.IsNotMissing(complexOperand))!;
```
ncipollina: **Potential silent fallthrough on scalar nullable `HasValue`.**

When `HasValue` is used on a *scalar* `Nullable<T>` (e.g., `where e.Score.HasValue`), `TryTranslateComplexPropertyForStructuralComparison` returns `null` because `FindComplexProperty` will fail on a plain scalar property.

The code then falls through to `TryGetMemberAccessChain(node, ...)`. `node` here is `e.Score.HasValue` — a `MemberExpression` whose `Expression` is `e.Score` (another `MemberExpression`) and whose own `Member.Name` is `"HasValue"`. Because `HasValue` is a `Nullable<T>` member, `IsNullableMember` will strip it from the names list. The chain walk continues up to `e.Score`, adding `"Score"`. So `names = ["Score"]`, and the result is `sqlExpressionFactory.Property("Score", bool)` — translating `e.Score.HasValue` as if it were `e.Score` itself (typed as `bool`).

This produces semantically incorrect SQL for queries like `Where(e => e.OptionalInt.HasValue)`. The translation silently produces `WHERE "optionalInt"` (a map attribute path), not `WHERE "optionalInt" IS NOT NULL AND "optionalInt" IS NOT MISSING`.

Consider either:
1. Detecting the scalar-nullable `HasValue` case in the `HasValue` branch and handling it correctly (translate the inner expression to a column and emit IS NOT NULL / IS NOT MISSING), or
2. Adding an explicit check after the `TryTranslateComplexPropertyForStructuralComparison` fallthrough so that a stray `HasValue` member name in the chain is rejected with `MemberAccessNotSupported` rather than silently mistyped.

If scalar `Nullable<T>.HasValue` is intentionally out of scope for this PR (focus is complex value types only), add a comment explaining why scalar nullable is excluded.

src/EntityFrameworkCore.DynamoDb/Query/Internal/DynamoSqlTranslatingExpressionVisitor.cs
```diff
@@ -303,7 +309,9 @@ private bool TryGetMemberAccessChain(
         while (true)
             if (sourceExpression is MemberExpression memberExpression)
             {
-                names.Add(memberExpression.Member.Name);
+                if (!IsNullableMember(memberExpression.Member, nameof(Nullable<int>.Value)))
+                    names.Add(memberExpression.Member.Name);
```
ncipollina, line 316: The omit-not-skip distinction is correct. For `x.OptionalAssociate.Value.Int` the chain traversal visits:
- `Int` → added (not a Nullable member)
- `Value` → skipped by this guard (is `Nullable<T>.Value`)
- `OptionalAssociate` → added

After `Reverse()` at line 332: `["OptionalAssociate", "Int"]` — the attribute path resolves to `"optionalAssociate"."int"`, which matches the expected SQL in the test. This is correct.

One nit: the `sourceExpression = memberExpression.Expression` assignment on the next line still walks *through* the stripped `Value` node to reach its parent, which is the right behavior — we continue walking the chain but don't record the `Value` segment in names.

src/EntityFrameworkCore.DynamoDb/Query/Internal/DynamoSqlTranslatingExpressionVisitor.cs
```diff
@@ -438,12 +444,11 @@ protected override Expression VisitMember(MemberExpression node)
...
         }
 
-        names.Reverse(); // Root-first order: e.g. [Profile, Address, City]
-
-        // Fast path: single-level direct property access — preserve existing behaviour exactly
+        // Fast path: single-level direct property access.
         if (names.Count == 1)
         {
-            if (rootEntityType?.FindProperty(node.Member.Name) is { } directProperty)
+            var memberName = names[0];
```
ncipollina, line 468: Good fix. The old code used `node.Member.Name` directly, which would have been wrong if the single-member chain was reached via an `EF.Property<T>()` call (where `node.Member.Name` and the resolved name differ). Using `names[0]` here is more consistent and correct regardless of how the chain was built.

tests/EntityFrameworkCore.DynamoDb.SpecificationTests/Query/Associations/ComplexProperties/ComplexPropertiesMiscellaneousDynamoTest.cs
```diff
@@ -0,0 +1,161 @@
...
+            typeof(ComplexPropertiesMiscellaneousDynamoTest));
+
+    public override Task Where_on_associate_scalar_property()
+        => base.Where_on_associate_scalar_property();
+
+    public override Task Where_on_optional_associate_scalar_property()
+        => base.Where_on_optional_associate_scalar_property();
+
+    public override Task Where_on_nested_associate_scalar_property()
+        => base.Where_on_nested_associate_scalar_property();
```
ncipollina: These three delegation-only overrides (`Where_on_associate_scalar_property`, `Where_on_optional_associate_scalar_property`, `Where_on_nested_associate_scalar_property`) satisfy `Check_all_tests_overridden` without verifying any SQL. If the goal is to confirm the tests pass without crashing, that is fine as a baseline. But it means the PartiQL generated for these cases is untested. Consider adding `AssertSql` baselines for them in a follow-up — at minimum for `Where_on_nested_associate_scalar_property` which exercises the multi-level chain that was just refactored.

src/EntityFrameworkCore.DynamoDb/Query/Internal/DynamoSqlTranslatingExpressionVisitor.cs
```diff
@@ -234,13 +238,13 @@ private static bool IsLiteralShape(Expression expression)
         {
             ConstantExpression => true,
             // Only allow field access — property getters may execute arbitrary user code.
-            MemberExpression { Member: System.Reflection.FieldInfo } memberExpression
-                => IsLiteralShape(memberExpression.Expression!),
+            MemberExpression { Member: System.Reflection.FieldInfo } memberExpression =>
+                IsLiteralShape(memberExpression.Expression!),
             UnaryExpression
             {
```
ncipollina, line 247: Style consistency cleanup noted. The trailing-comma removals across switch expression arms (this hunk and several others) are consistent with the project's .editorconfig and make diffs cleaner. Good housekeeping.

docs/querying/filtering.md
```diff
@@ -34,6 +34,17 @@ DynamoDB has two distinct representations of "no value" for an attribute: a `NUL
...
+Nullable complex value objects use the same null/missing rules. `HasValue` checks that the complex map attribute is present and not `NULL`; accessing `.Value.Property` translates to a nested attribute path.
+
+```csharp
+var active = await db.Customers
+    .Where(c => c.Profile.HasValue && c.Profile.Value.Address.ZipCode == "98101")
+    .ToListAsync(cancellationToken);
+
+// WHERE "Profile" IS NOT NULL AND "Profile" IS NOT MISSING
+//   AND "Profile"."Address"."ZipCode" = ?
+```
```
ncipollina, line 46: Minor doc nit: the example uses `c.Profile.HasValue` on a `Customers` entity with a `Profile` complex property, but the spec fixture and the actual feature is demonstrated on `ValueRootEntity.OptionalAssociate`. The example is illustrative-only (not a verbatim copy of test data), which is fine — but make sure if someone copies this snippet to a real project, the `Profile` property is actually a nullable complex type (`ComplexProperty(...).IsRequired(false)`), not a reference-type complex property. A one-line note clarifying this must be a struct-based value type wrapped in `Nullable<T>` would remove any ambiguity for readers.

src/EntityFrameworkCore.DynamoDb/Query/Internal/DynamoSqlTranslatingExpressionVisitor.cs
ncipollina, line 316: `TryTranslateComplexPropertyForStructuralComparison` (called from the `HasValue` branch above) correctly returns `null` when the chain doesn't resolve through complex properties — `FindComplexProperty` returns `null` for scalars so the function short-circuits. The `HasValue` branch correctly checks this before generating SQL. The only subtlety is what happens for scalar nullable after the `null` return — noted in the comment above on `VisitMember`.

ncipollina (changes requested):
## Review: fix(query): translate nullable complex value members

Overall this is a solid, well-scoped addition. The core translation logic is correct for the intended case (nullable complex value-type properties), `IsNullableMember` is properly guarded against false positives, the `names[0]` fast-path fix is an improvement over the old `node.Member.Name` direct access, and the spec-test baseline coverage is welcome. The style-only trailing-comma cleanup is consistent with the project's editorconfig.

### Blocking issue

**Scalar nullable `HasValue` silently produces wrong SQL.**

The `VisitMember` `HasValue` branch is a conditional — if `TryTranslateComplexPropertyForStructuralComparison` returns `null` (which it does for scalar `Nullable<T>` properties, since `FindComplexProperty` returns `null` for scalars), execution falls through to `TryGetMemberAccessChain(node, ...)`.

At that point `node` is `e.Score.HasValue`. The chain walker visits `HasValue` first, strips it (because `IsNullableMember` returns true), then visits `Score`, adds it. `names = ["Score"]` after reverse. The fast path returns `sqlExpressionFactory.Property("Score", bool)` — a map attribute path typed as `bool`, not `IS NOT NULL AND IS NOT MISSING`.

A query `Where(e => e.OptionalInt.HasValue)` would silently emit `WHERE "optionalInt"` instead of `WHERE "optionalInt" IS NOT NULL AND "optionalInt" IS NOT MISSING`. This is semantically incorrect.

**Fix options (either works):**
1. In the `HasValue` branch, when `TryTranslateComplexPropertyForStructuralComparison` returns `null`, also attempt to translate `nullableExpression` as a plain scalar column and, if that succeeds, emit the same `IS NOT NULL AND IS NOT MISSING` pair.
2. After the `HasValue` branch falls through to `TryGetMemberAccessChain`, detect that the top-level `node.Member` was a stripped `HasValue` member (e.g., check `names.Count < original chain length via a HasValue ancestor check`) and reject with `MemberAccessNotSupported`.

If scalar nullable `HasValue` is intentionally out of scope for this PR, add an explicit comment in the `HasValue` branch noting that and ensure a future issue tracks it.

### Non-blocking notes

- The three delegation-only overrides (`Where_on_associate_scalar_property`, `Where_on_optional_associate_scalar_property`, `Where_on_nested_associate_scalar_property`) pass through to base without `AssertSql`. The nested one exercises the refactored multi-level chain path — worth adding a SQL baseline in a follow-up.
- The `filtering.md` doc example is clear; a one-line clarification that `Profile` must be a struct-based `Nullable<T>` complex property (not a reference-type optional) would help users avoid a confusing setup mistake.

### Summary

| | Finding |
|---|---|
| Blocking | Scalar `Nullable<T>.HasValue` translates to wrong attribute path due to fallthrough in `VisitMember` |
| Non-blocking | 3 delegation-only test overrides have no `AssertSql` baseline |
| Non-blocking | Doc example could clarify struct-vs-reference-type requirement |
| Positive | `IsNullableMember` guard is correct and tight |
| Positive | `names[0]` fast-path fix is an improvement |
| Positive | Trailing-comma cleanup is consistent with project style |

src/EntityFrameworkCore.DynamoDb/Query/Internal/DynamoSqlTranslatingExpressionVisitor.cs
j-d-ha: Fixed. `Nullable<T>.HasValue` now translates the inner nullable operand first, falling back from complex-map path to scalar column translation, then emits `IS NOT NULL AND IS NOT MISSING`. Added `Nullable_enum_has_value` coverage asserting scalar nullable SQL.

tests/EntityFrameworkCore.DynamoDb.SpecificationTests/Query/Associations/ComplexProperties/ComplexPropertiesMiscellaneousDynamoTest.cs
j-d-ha: Added `AssertSql` baselines for all three overrides, including the nested complex path case.

docs/querying/filtering.md
j-d-ha: Clarified this in `filtering.md`: nullable complex `HasValue` examples require a struct-based value object (`Profile?`) configured optional, not a nullable reference-type complex property.

src/EntityFrameworkCore.DynamoDb/Query/Internal/DynamoSqlTranslatingExpressionVisitor.cs
j-d-ha: Thanks, agreed — keeping the `Nullable<>` declaring-type guard as-is.

src/EntityFrameworkCore.DynamoDb/Query/Internal/DynamoSqlTranslatingExpressionVisitor.cs
j-d-ha: Agreed. Kept traversal behavior: walk through `Nullable<T>.Value` while omitting that segment from the DynamoDB path.

src/EntityFrameworkCore.DynamoDb/Query/Internal/DynamoSqlTranslatingExpressionVisitor.cs
j-d-ha: Agreed. Keeping `names[0]` so resolved chain names are used consistently for direct property access.

src/EntityFrameworkCore.DynamoDb/Query/Internal/DynamoSqlTranslatingExpressionVisitor.cs
j-d-ha: Thanks. No code change needed here.

src/EntityFrameworkCore.DynamoDb/Query/Internal/DynamoSqlTranslatingExpressionVisitor.cs
j-d-ha: Covered by the scalar nullable `HasValue` fix above: when complex-path translation returns null, the branch now translates the inner scalar operand and emits the null/missing predicate instead of falling through.

ncipollina (approved):
Blocking issue resolved. The `?? TranslateInternal(nullableExpression)` fallback correctly handles scalar nullable `HasValue` (e.g. `b.Enum.HasValue`) — the complex path is tried first, scalar path second, and both produce `IS NOT NULL AND IS NOT MISSING`. The new `Nullable_enum_has_value` test pins the baseline. Docs update clarifying struct-based vs reference-type scope is a nice touch. Approved.

PR #263, opened 2026-06-24
fix(query): preserve discriminator predicates in type filters

j-d-ha:
## Summary

Preserves discriminator predicate metadata when translating C# type filters so downstream query generation can handle them correctly. Documents that `OfType<T>()`, `is`, and `GetType()` type filtering require active discriminator metadata when shared-table discrimination has been disabled.

## Changes

- Wrap translated `is` predicates and `GetType()` equality predicates in `SqlDiscriminatorPredicateExpression`.
- Preserve negation behavior for `GetType() != typeof(...)` through the wrapped discriminator predicate.
- Clarify shared-table discriminator limits in `docs/limitations.md` and `docs/modeling/single-table-design.md`.

## Validation

- `dotnet build`

## Release Notes

- Type filters over shared-table inheritance now require active discriminator metadata; models using `HasNoDiscriminator()` should use key predicates or concrete `DbSet` queries when keys already isolate types.


ncipollina (changes requested):
## Review: fix(query) — preserve discriminator predicates in type filters

This is a substantial, well-structured addition. The core translation work (OfType, `is`, GetType()) is solid and the spec test coverage is comprehensive. Several correctness and design points to address before merging.

### Must Fix

**1. Cross-root assignability guard is over-constrained (both visitors)**
The double-sided `GetRootType()` check rejects valid same-hierarchy queries. If `sourceEntityType` is `Animal` (root) and `targetEntityType` is `Kiwi` (derived), `targetEntityType.GetRootType() == Animal`, which *is* assignable from `sourceEntityType.GetRootType() == Animal`. So the second clause always passes for valid same-hierarchy cases, meaning the only real guard is the first. But the first clause can reject when the source is a non-root base (e.g., `Bird` navigating to `Kiwi`) if the roots differ across mapped tables. The two-clause pattern should be `sourceRoot == targetRoot`; the current form is confusing and has a latent bug for multi-table mixed hierarchies.

**2. `VisitTypeBinary` ignores `TypeAs` / other node types silently**
`ExpressionType.TypeAs` is a valid `TypeBinaryExpression` node type (C# `as` operator). The current implementation returns `NotTranslatedExpression` for it, which is correct, but the `node.NodeType == ExpressionType.TypeIs` check burns through both cases in one ternary, making it harder to see intent. A named constant or early-return guard would be clearer.

**3. `TranslateCollectionAny` — `IsNotMissing` is semantically wrong for empty collections**
`IsNotMissing` checks attribute existence, but DynamoDB can store an empty list that *is present* but has zero elements. A stored `[]` passes `IS NOT MISSING` but `.Any()` should return false. The correct translation would be a size check (e.g., `size(attr) > 0`) if supported, or fall back to `NotTranslatedExpression` for now. This is a correctness bug.

**4. `NoSyncTest(expectSyncFailure: false)` — silent success when exception is not thrown**
The original `NoSyncTest` always called `Assert.Fail` when sync succeeded unexpectedly. The new `expectSyncFailure = false` path returns silently even when the sync call threw a *different* exception type. That swallows unexpected failures for the six callers that pass `false`. The catch clause should remain specific: only catch `IsExpectedSyncQueryFailure` exceptions, and re-throw anything else regardless of the flag.

### Suggestions

**5. `CreateFalsePredicate()` — `1 = 0` is fragile**
A literal `1 = 0` may or may not be valid PartiQL. A dedicated `SqlFalseExpression` or a known-false constant expression would be safer. At minimum, add a unit test that verifies the false predicate serializes to valid PartiQL.

**6. Redundant discriminator predicate in `OfType` + entity-type root filter**
`Can_use_of_type_bird_with_projection` shows `WHERE ("discriminator" = 'Eagle' OR "discriminator" = 'Kiwi') AND ("discriminator" = 'Eagle' OR "discriminator" = 'Kiwi')`. The root entity materializer applies its own discriminator predicate, and `TranslateOfType` appends another. EF Cosmos deduplicates this. The query still executes correctly (it's AND of identical predicates), but it's wasteful and looks like a bug in generated output. Worth a follow-up issue.

**7. `#if NET11_0_OR_GREATER` in `ComplianceDynamoTest.cs` — guarded `using` without guarded type reference**
`InheritanceQueryTestBase<>` is yielded unconditionally on EF10 via the EF10 namespace import at the top of the file. The `#if NET11_0_OR_GREATER` guard on the `using` is inverted from what `InheritanceQueryDynamoTest.cs` does (which guards with `#if NET10_0` for the old namespace). Double-check that EF10 can resolve `InheritanceQueryTestBase<>` from `Microsoft.EntityFrameworkCore.Query` as imported by `ComplianceDynamoTest.cs`.

### Positives

✅ The `TryUnwrapGetTypeComparison` with bidirectional argument order check (`|| TryUnwrap(right, left, ...)`) correctly handles both `GetType() == typeof(T)` and `typeof(T) == GetType()` forms.

✅ Wrapping discriminator predicates in `SqlDiscriminatorPredicateExpression` before passing to `Not()` is clean — it preserves the predicate's identity for downstream inspection.

✅ The `exact` parameter that selects `GetConcreteDerivedTypesInclusive` vs. `[targetEntityType]` is a clear, non-duplicated design for `is` vs. `GetType()` semantics.

✅ The `OfTypeRequiresDiscriminatorPredicate` error string clearly distinguishes the no-discriminator-metadata failure from the general unsupported case.

✅ Promoting `InheritanceQueryTestBase` from the Future table to Implemented in `docs/spec-test-coverage.md` with accurate method counts is the right housekeeping.

✅ The `expectSyncFailure` parameter solves the real problem (tests that succeed on sync path but EF throws for unrelated reasons). The fix direction is right; only the exception re-throw gap needs closing.

src/EntityFrameworkCore.DynamoDb/Query/Internal/DynamoSqlTranslatingExpressionVisitor.cs
```diff
@@ -278,6 +294,142 @@ or QueryParameterExpression
...
+        Expression source,
+        Type targetClrType,
+        bool exact)
+    {
+        var sourceEntityType = ResolveRootEntityType(source);
+        var targetEntityType = model.FindEntityType(targetClrType);
+        if (sourceEntityType is null || targetEntityType is null)
+            return null;
+
+        if (!sourceEntityType.GetRootType().IsAssignableFrom(targetEntityType)
```
ncipollina: ⚠️ **Cross-root guard — the double-sided form is confusing and fragile**

```csharp
if (!sourceEntityType.GetRootType().IsAssignableFrom(targetEntityType)
    || !targetEntityType.GetRootType().IsAssignableFrom(sourceEntityType))
```

This correctly rejects cross-hierarchy types, but the intent is simpler: just require the roots to be the same entity type. The two-clause form passes for same-hierarchy pairs because each root is assignable from the other, but it reads as if two distinct invariants are being enforced. A forthcoming multi-root scenario could silently slip through if one clause is accidentally satisfied.

Prefer the explicit intent:
```csharp
if (sourceEntityType.GetRootType() != targetEntityType.GetRootType())
    return exact ? CreateFalsePredicate() : null;
```
This is cleaner, faster (one equality check vs. two `IsAssignableFrom` calls), and makes the guard self-documenting.

src/EntityFrameworkCore.DynamoDb/Query/Internal/DynamoSqlTranslatingExpressionVisitor.cs
```diff
@@ -1392,6 +1556,19 @@ and not SqlParameterExpression
...
+    private Expression TranslateCollectionAny(MethodCallExpression node)
+    {
+        var sourceExpression = StripAsQueryable(node.Arguments.ElementAtOrDefault(0));
+        if (sourceExpression is null)
+            return QueryCompilationContext.NotTranslatedExpression;
+
+        if (TranslateInternal(sourceExpression) is not SqlExpression source)
+            return QueryCompilationContext.NotTranslatedExpression;
+
+        return sqlExpressionFactory.IsNotMissing(source);
```
ncipollina: ❌ **`IsNotMissing` is semantically wrong for `.Any()` on a primitive collection**

`IS NOT MISSING` checks attribute existence — it returns `true` for a stored empty list `[]`. The LINQ `.Any()` semantic is "has at least one element", so `[].Any()` should return `false`.

DynamoDB PartiQL supports `size(attr) > 0` for this check. If `size()` isn't already translated by the provider, the correct fallback is `QueryCompilationContext.NotTranslatedExpression` — don't silently return the wrong predicate.

```csharp
// Current — wrong for empty collections:
return sqlExpressionFactory.IsNotMissing(source);

// Correct if size() is translatable:
return sqlExpressionFactory.GreaterThan(
    sqlExpressionFactory.Function("size", [source], typeof(int)),
    sqlExpressionFactory.Constant(0, typeof(int)));

// Or, safe fallback:
```

tests/EntityFrameworkCore.DynamoDb.SpecificationTests/TestUtilities/DynamoTestHelpers.cs
```diff
@@ -40,15 +40,24 @@ public void NoSyncTest(Action testCode)
         Assert.Fail("Expected DynamoDB sync query failure.");
     }
 
-    /// <summary>Runs an async-aware test and verifies that sync DynamoDB query execution fails.</summary>
-    public async Task NoSyncTest(bool async, Func<bool, Task> testCode)
+    /// <summary>Runs an async-aware test and optionally verifies that sync DynamoDB query execution fails.</summary>
+    public async Task NoSyncTest(
+        bool async,
+        Func<bool, Task> testCode,
+        bool expectSyncFailure = true)
```
ncipollina, line 47: ⚠️ **`expectSyncFailure = false` silently swallows unexpected exceptions**

The current code catches `InvalidOperationException` where `IsExpectedSyncQueryFailure` matches, then `return`s. After the catch, it checks `!async && expectSyncFailure` to decide whether to `Assert.Fail`. But if `expectSyncFailure == false` and the sync path throws a *different* exception (e.g., `NullReferenceException`), the catch block does **not** match, so the exception propagates — that part is fine. However, if the sync path succeeds (no exception) and `expectSyncFailure == false`, we return silently — that's the intent.

The real issue is: what if `expectSyncFailure == false` but the sync path throws `InvalidOperationException` with `IsExpectedSyncQueryFailure == true`? The catch fires, `return`s early, and the test passes — but we expected it to succeed without any exception. Consider whether the early `return` in the catch should only fire when `expectSyncFailure == true`:

```csharp
catch (InvalidOperationException exception) when (!async && IsExpectedSyncQueryFailure(exception))
{
    if (expectSyncFailure)
        return;
    // If we did NOT expect sync failure but got it anyway, re-throw
    throw;
}

if (!async && expectSyncFailure)
    Assert.Fail("Expected DynamoDB sync query failure.");
```

This ensures `expectSyncFailure = false` tests see any unexpected sync-path exceptions rather than swallowing them.

src/EntityFrameworkCore.DynamoDb/Query/Internal/DynamoSqlTranslatingExpressionVisitor.cs
```diff
@@ -278,6 +294,142 @@ or QueryParameterExpression
         }
     }
 
+    /// <inheritdoc />
+    protected override Expression VisitTypeBinary(TypeBinaryExpression node)
```
ncipollina, line 298: 🤔 **`VisitTypeBinary` — add explicit `TypeAs` branch for clarity**

`TypeBinaryExpression` covers both `TypeIs` (`x is T`) and `TypeAs` (`x as T`). The current ternary handles `TypeAs` by falling through to `NotTranslatedExpression`, which is correct. But an explicit branch documents intent and prevents a future reader from wondering whether `TypeAs` was intentionally skipped:

```csharp
protected override Expression VisitTypeBinary(TypeBinaryExpression node)
{
    if (node.NodeType != ExpressionType.TypeIs)

    return TryCreateDiscriminatorPredicate(node.Expression, node.TypeOperand, false) is { } predicate
        ? new SqlDiscriminatorPredicateExpression(predicate)
        : QueryCompilationContext.NotTranslatedExpression;
}
```

Minor, but worth a single-line guard for readability.

src/EntityFrameworkCore.DynamoDb/Query/Internal/DynamoQueryableMethodTranslatingExpressionVisitor.cs
```diff
@@ -749,7 +795,63 @@ private bool RequiresDiscriminatorPredicate(IEntityType entityType)
...
+            {
+                StructuralType: IEntityType sourceEntityType
+            })
+            return UnsupportedOperator(
+                nameof(Queryable.OfType),
+                DynamoStrings.OfTypeNotSupportedYet);
+
+        var targetEntityType = QueryCompilationContext.Model.FindEntityType(resultType);
+        if (targetEntityType is null
+            || !sourceEntityType.GetRootType().IsAssignableFrom(targetEntityType)
```
ncipollina: ⚠️ **Same double-sided root check as `DynamoSqlTranslatingExpressionVisitor` — same simplification applies**

```csharp
// Current
if (targetEntityType is null
    || !sourceEntityType.GetRootType().IsAssignableFrom(targetEntityType)
    || !targetEntityType.GetRootType().IsAssignableFrom(sourceEntityType))

// Simpler and equivalent
if (targetEntityType is null
    || sourceEntityType.GetRootType() != targetEntityType.GetRootType())
```

See the corresponding comment in `DynamoSqlTranslatingExpressionVisitor`. Both methods have the same pattern — if you change one, change both.

src/EntityFrameworkCore.DynamoDb/Query/Internal/DynamoSqlTranslatingExpressionVisitor.cs
```diff
@@ -278,6 +294,142 @@ or QueryParameterExpression
...
+            || TryUnwrapGetTypeComparison(node.Right, node.Left, out instance, out type))
+        {
+            var predicate = TryCreateDiscriminatorPredicate(instance, type, true);
+            if (predicate is null)
+                return null;
+
+            var discriminatorPredicate = new SqlDiscriminatorPredicateExpression(predicate);
+
+            return node.NodeType == ExpressionType.NotEqual
+                ? sqlExpressionFactory.Not(discriminatorPredicate)
```
ncipollina, line 321: ✅ **Good: wrapping in `SqlDiscriminatorPredicateExpression` before negating preserves the predicate's identity**

Wrapping first and then calling `Not()` on the wrapper (rather than negating the inner SQL first and then wrapping) means any downstream code that inspects `SqlDiscriminatorPredicateExpression` nodes still sees the affirmative predicate inside, with the negation on the outside. This makes the expression tree easier to inspect, rewrite, and serialize. Nice design choice.

tests/EntityFrameworkCore.DynamoDb.SpecificationTests/Query/InheritanceQueryDynamoTest.cs
```diff
@@ -0,0 +1,583 @@
...
+    public override Task Can_use_of_type_bird(bool async)
+        => DynamoTestHelpers.Instance.NoSyncTest(async, a => base.Can_use_of_type_bird(a));
+
+    [ConditionalTheory(Skip = SkipReason.OrderedResultSetNotSupported)]
+    public override Task Can_use_of_type_bird_predicate(bool async)
+        => DynamoTestHelpers.Instance.NoSyncTest(
+            async,
+            a => base.Can_use_of_type_bird_predicate(a));
+
+    public override Task Can_use_of_type_bird_with_projection(bool async)
```
ncipollina, line 152: 🤔 **Redundant discriminator predicate in generated SQL — worth a follow-up issue**

The expected SQL for this test contains:
```sql
WHERE ("discriminator" = 'Eagle' OR "discriminator" = 'Kiwi') AND ("discriminator" = 'Eagle' OR "discriminator" = 'Kiwi')
```

The predicate is `AND`-ed with itself. This happens because the entity materializer applies a root-type discriminator predicate independently from the one `TranslateOfType` adds. The query still returns correct results (AND of equal predicates), but it wastes work at the PartiQL level.

The test expectation is correctly documenting the current behavior, which is the right call for a spec test. But I'd suggest opening a follow-up issue to deduplicate discriminator predicates in `SelectExpression.ApplyPredicate` — EF Cosmos does this.

tests/EntityFrameworkCore.DynamoDb.SpecificationTests/Query/InheritanceQueryDynamoTest.cs
```diff
@@ -0,0 +1,583 @@
...
+                await base.Can_use_of_type_kiwi(a);
+                AssertSql(
+                    """
+                    SELECT "id", "countryId", "discriminator", "name", "species", "isFlightless", "foundOn"
+                    FROM "Animals"
+                    WHERE "discriminator" = 'Kiwi' AND ("discriminator" = 'Eagle' OR "discriminator" = 'Kiwi')
+                    """);
+            });
+
+    public override Task Can_use_backwards_of_type_animal(bool async)
```
ncipollina, line 184: ✅ **Good: `Can_use_backwards_of_type_animal` exercises the identity/base-type no-op path**

This test calls `OfType<Animal>` on a `DbSet<Kiwi>` — the target is a base type of the source, so no discriminator predicate is needed and the source is returned unchanged. The expected SQL (`WHERE "discriminator" = 'Kiwi'`) comes purely from the root entity materializer. This validates the `IsOfTypeIdentityOrBaseType` short-circuit correctly.

src/EntityFrameworkCore.DynamoDb/Query/Internal/DynamoSqlTranslatingExpressionVisitor.cs
```diff
@@ -278,6 +294,142 @@ or QueryParameterExpression
...
+        bool exact)
+    {
+        var sourceEntityType = ResolveRootEntityType(source);
+        var targetEntityType = model.FindEntityType(targetClrType);
+        if (sourceEntityType is null || targetEntityType is null)
+            return null;
+
+        if (!sourceEntityType.GetRootType().IsAssignableFrom(targetEntityType)
+            || !targetEntityType.GetRootType().IsAssignableFrom(sourceEntityType))
+            return exact ? CreateFalsePredicate() : null;
```
ncipollina, line 382: 🤔 **`CreateFalsePredicate` — verify `1 = 0` is valid PartiQL**

The literal `1 = 0` generates a constant-false predicate. DynamoDB PartiQL may accept this (it is standard SQL), but if you hit any execution errors on it, consider using `false` directly as a Boolean constant if the factory supports it, or restructure to return `NotTranslatedExpression` and let the caller propagate an unsupported operation. Either way, a unit test verifying that `TryCreateDiscriminatorPredicate` with a cross-hierarchy type produces valid serialized PartiQL (not just a non-null expression) would close this risk.

ncipollina, line 382: ✅ **Good: single `TryCreateDiscriminatorPredicate(exact: bool)` covers both `is` and `GetType()` without duplication**

The `exact` parameter elegantly differentiates:
- `x is T` → `exact: false` → all concrete derived types inclusive (polymorphic match)  
- `x.GetType() == typeof(T)` → `exact: true` → only the named concrete type itself

This avoids two near-duplicate methods and makes the difference between the two semantics immediately visible at call sites. Good design.

src/EntityFrameworkCore.DynamoDb/Query/Internal/DynamoQueryableMethodTranslatingExpressionVisitor.cs
```diff
@@ -749,7 +795,63 @@ private bool RequiresDiscriminatorPredicate(IEntityType entityType)
...
+                targetEntityType,
+                projectionBindingExpression,
+                false));
+    }
+
+    private static bool IsOfTypeIdentityOrBaseType(
+        IEntityType sourceEntityType,
+        IEntityType targetEntityType)
+        => sourceEntityType == targetEntityType
+            || targetEntityType.IsAssignableFrom(sourceEntityType);
```
ncipollina: 🤔 **`IsOfTypeIdentityOrBaseType` — first clause is subsumed by the second**

```csharp
=> sourceEntityType == targetEntityType
    || targetEntityType.IsAssignableFrom(sourceEntityType);
```

If `sourceEntityType == targetEntityType`, then `targetEntityType.IsAssignableFrom(sourceEntityType)` is trivially true (every type is assignable from itself). The first clause is redundant. This is harmless but can be simplified to:

```csharp
=> targetEntityType.IsAssignableFrom(sourceEntityType);
```

The identity case is already covered.

ncipollina (approved):
All three must-fix items from the previous review have been correctly addressed:

- ✅ `TranslateCollectionAny` now uses `size(attr) > 0` — semantically correct for primitive collection `.Any()`
- ✅ `NoSyncTest` re-throws when `expectSyncFailure == false` — no more silent swallowing
- ✅ Double-sided `IsAssignableFrom` replaced with clean `!=` equality check

Happy to approve. 🎉

PR #266, opened 2026-06-26
feat!: enable Dynamo discriminators by default

j-d-ha:
## Summary

DynamoDB entity mappings now get a discriminator by default, including single-entity tables. This makes persisted items self-describing, validates discriminator values and key-name collisions earlier, and documents migration/backfill or opt-out paths for existing tables.

## Changes

- Enable default discriminator metadata per table group unless explicitly opted out with `HasNoDiscriminator()`.
- Extend model validation for discriminator value requirements, duplicate values, attribute-name mismatch, and PK/SK collisions.
- Update integration seed mappers and PartiQL baselines to include projected/persisted `$type`.
- Refresh docs and XML docs for default `$type`, query projection behavior, backfill guidance, and opt-out behavior.

## Validation

- `dotnet test tests/EntityFrameworkCore.DynamoDb.Tests/EntityFrameworkCore.DynamoDb.Tests.csproj` — passed: 737 passed, 2 skipped.
- `dotnet test tests/EntityFrameworkCore.DynamoDb.IntegrationTests/EntityFrameworkCore.DynamoDb.IntegrationTests.csproj` — passed: 443 passed, 2 skipped.
- `task docs:build` — passed.
- `git diff --check origin/main...HEAD` — clean.

## Breaking Changes

- What changed: mapped entities now persist/read a default discriminator attribute (`$type`) unless discriminator mapping is explicitly disabled.
- Previous behavior: single-entity tables generally did not require or write `$type`.
- New behavior: new items include `$type`, projections include `$type`, and validation rejects discriminator/key collisions.
- Migration/action needed: backfill existing tables with the expected `$type` value or opt out with `HasNoDiscriminator()` where discriminator metadata is not desired.

## Release Notes

Default DynamoDB discriminators are now enabled for mapped entities. Existing tables may need a `$type` backfill, or explicit `HasNoDiscriminator()` configuration to preserve old item shapes.

## Notes for Reviewers

Parent-orchestrated review loop ran 3 rounds. Remaining reviewer notes are docs wording polish only: `docs/modeling/single-table-design.md` can be tightened from “when discrimination is active” to “when type filtering is required”, and internal comments can be aligned later.


src/EntityFrameworkCore.DynamoDb/Infrastructure/Internal/DynamoModelValidator.cs
```diff
@@ -518,30 +522,22 @@ private static void ValidateDiscriminatorMappingGroup(
...
 
-        // Mixed state (some with discriminator, some without) is normalized by
-        // DynamoDiscriminatorConvention before finalization — if any root in the group has an
-        // explicit no-discriminator signal, the convention propagates HasNoDiscriminator() to the
-        // entire group. By the time validation runs, all concrete types either have a discriminator
-        // property or none do, so concreteTypesWithDiscriminator.Count == concreteTypes.Count here.
-
         var first = concreteTypesWithDiscriminator[0];
         var expectedDiscriminatorProperty = first.FindDiscriminatorProperty()
             ?? throw new InvalidOperationException(
```
ncipollina, line 527: ⚠️ **Stale wording in secondary-loop error message**

The error message for a missing discriminator property on non-`first` entity types still says `"shared DynamoDB table"` and includes the old multi-type guidance sentence, even though the outer condition (`concreteTypesWithDiscriminator.Count == 0`) now handles the single-type-no-discriminator bypass differently. For consistency with the other updated messages, drop `"shared"` and remove the requirement sentence:

```csharp
// current (inside the orderedConcreteTypes loop):
$"Entity type '{entityType.DisplayName()}' is mapped to shared DynamoDB table '{tableName}' but does not define a discriminator property. "
+ "Shared-table mappings with multiple entity types require a discriminator property."

// suggested:
$"Entity type '{entityType.DisplayName()}' is mapped to DynamoDB table '{tableName}' but does not define a discriminator property."
```

The other two updated messages in this file already dropped `"shared"` — this one was missed.

ncipollina (commented):
## Review Summary

This PR makes discriminators the universal default for all mapped entity types — a clean, consistent design choice that aligns with how most DynamoDB single-table patterns evolve over time. The convention simplification, validator extension, and comprehensive integration baseline updates all look well-executed. A few targeted issues are noted in the inline comments, the most actionable being the stale error message wording in the secondary loop, a misleading integration test name, and a small gap in the positive-path unit test coverage for single-type tables.

**Disposition: Approve with suggestions** — no blocking correctness issues; the inline items are improvements worth folding in before merge if the author agrees.

tests/EntityFrameworkCore.DynamoDb.IntegrationTests/SharedTable/DiscriminatorQueryTests.cs
```diff
@@ -87,7 +87,7 @@ public async Task SingleTypeTable_DoesNotInjectDiscriminatorPredicate()
```
ncipollina, line 87: 💡 **Test name is now misleading**

`SingleTypeTable_DoesNotInjectDiscriminatorPredicate` was accurate before this PR because single-type tables had no `$type` in the SELECT at all. After the change the SELECT now includes `$type` — only the WHERE predicate injection is suppressed. Consider renaming to better reflect the post-change semantics, for example:

```csharp
public async Task SingleTypeTable_ProjectsDiscriminator_ButDoesNotInjectWherePredicate()
```

This makes it clear the column appears in projections but type-filtering is not added to the WHERE clause.

tests/EntityFrameworkCore.DynamoDb.Tests/Metadata/DiscriminatorValidationTests.cs
```diff
@@ -379,4 +428,43 @@ public void SharedTableMultipleTypes_WhenDiscriminatorNameCollidesWithSortKey_Th
             .Throw<InvalidOperationException>()
             .WithMessage("*collides with the sort key attribute name*");
     }
+
+    [Fact(Timeout = TestConfiguration.DefaultTimeout)]
+    public void SingleType_WithMissingDiscriminatorValue_Throws()
```
ncipollina, line 504: 💡 **Missing positive-path test for single-type table**

The new `SingleType_*` tests all cover failure cases. There is no test confirming that a plain single-type table with the convention-configured discriminator passes validation. Adding one would close the coverage gap and make the happy-path intent explicit:

```csharp
[Fact(Timeout = TestConfiguration.DefaultTimeout)]
public void SingleType_WithConventionDiscriminator_IsValid()
{
    var client = MockClient();
    // No explicit discriminator config — convention sets $type automatically
    using var context = new SingleTypeConventionContext(BuildOptions<SingleTypeConventionContext>(client));

    var act = () => context.Model;

    act.Should().NotThrow();
}
```

Where `SingleTypeConventionContext` maps a single entity to its own table with only `HasPartitionKey`/`HasSortKey` set. This mirrors what `SharedTableMultipleTypes_WithConventionDiscriminator_IsValid` does for the shared-table case.

src/EntityFrameworkCore.DynamoDb/Infrastructure/Internal/DynamoModelValidator.cs
```diff
@@ -518,30 +522,22 @@ private static void ValidateDiscriminatorMappingGroup(
...
+                $"Entity type '{first.DisplayName()}' is mapped to DynamoDB table '{tableName}' but does not define a discriminator value.");
 
         var expectedDiscriminatorAttributeName = expectedDiscriminatorProperty.GetAttributeName();
 
         var valuesByEntityType = new Dictionary<object, IEntityType>(
             expectedDiscriminatorProperty.GetKeyValueComparer());
         valuesByEntityType[expectedDiscriminatorValue] = first;
 
-        foreach (var entityType in concreteTypesWithDiscriminator.Skip(1))
+        foreach (var entityType in orderedConcreteTypes.Where(entityType => entityType != first))
```
ncipollina, line 543: ❓ **Loop now iterates all concrete types, not just those with a discriminator**

The old loop was `concreteTypesWithDiscriminator.Skip(1)` — it only touched types that already had a discriminator property and relied on the removed comment ("By the time validation runs, all concrete types either have a discriminator property or none do") as the invariant.

The new loop is `orderedConcreteTypes.Where(entityType => entityType != first)` — it iterates ALL types except `first`, regardless of whether they have a discriminator. If any type has no discriminator it hits the inner `FindDiscriminatorProperty() ?? throw` guard, which is correct behaviour.

This is fine in practice because `DynamoDiscriminatorConvention` enforces that either the whole group has discriminators or the whole group is opted out before validation runs. But the removed comment was the only documentation of that invariant. A brief inline note here would help future readers understand why a type missing its discriminator can still appear in this loop without causing a false positive:

```csharp
// If any concrete type has a discriminator, they all should — the convention normalises the
// whole group before model finalization. Any type missing it here is a misconfiguration and
// will throw on the FindDiscriminatorProperty() guard below.
foreach (var entityType in orderedConcreteTypes.Where(entityType => entityType != first))
```

tests/EntityFrameworkCore.DynamoDb.IntegrationTests/PkSkTable/Infra/Mappers.cs
```diff
@@ -15,4 +15,7 @@ internal static List<Dictionary<string, AttributeValue>> ToItems(List<PkSkItem>
 
     internal static List<PkSkItem> FromItems(List<Dictionary<string, AttributeValue>> items)
         => items.Select(FromItem).ToList();
+
```
ncipollina, line 18: ✅ **Correct use of AfterToItem hook**

Using `AfterToItem` to inject `$type` into seed data is the right approach — it keeps the mapper partial class clean and mirrors the actual runtime write path. One thing to watch: `nameof(PkSkItem)` and the convention's `entityType.ShortName()` are equivalent for top-level classes, so the seed values will match what the provider writes. If any entity type were a nested class these could diverge (`ShortName()` typically returns just the simple name), but there are no nested entity types in the integration fixtures so this is fine as-is.

ncipollina, line 18: ✅ **Correct use of AfterToItem hook**

Using `AfterToItem` to inject `$type` into seed data is the right approach — it keeps the mapper partial class clean and mirrors the actual runtime write path. One detail to note: `nameof(PkSkItem)` and the convention `entityType.ShortName()` are equivalent for top-level classes, so the seed values will always match what the provider writes. If any entity type were a nested class these could diverge, but there are no nested entity types in the integration fixtures so this is safe as-is.

docs/limitations.md
```diff
@@ -394,22 +394,19 @@ update the token value before calling `SaveChangesAsync`.
...
-type filters cannot be translated.
-
-When discriminator metadata is enabled, the following constraints are validated at startup:
+A discriminator is configured by default for every mapped entity type. Shared-table mappings and
+inheritance rely on it for type-safe filtering and materialization. The following constraints are
+validated at startup:
 
 - Discriminator values must be unique within the table group.
 - All entity types in the group must use the same discriminator attribute name.
 - The discriminator attribute name must not collide with the partition key or sort key attribute
```
ncipollina: ✅ **Migration guidance is clear and prominent**

The new sentence — "Existing tables without this attribute must be backfilled or mapped with `HasNoDiscriminator()`." — is exactly the right escape-hatch callout for users upgrading from a version where single-type tables had no discriminator. Putting it directly in the constraints section means it will not be missed.

tests/EntityFrameworkCore.DynamoDb.Tests/Metadata/DiscriminatorValidationTests.cs
```diff
@@ -379,4 +428,43 @@ public void SharedTableMultipleTypes_WhenDiscriminatorNameCollidesWithSortKey_Th
...
+        var act = () => context.Model;
+
+        act
+            .Should()
+            .Throw<InvalidOperationException>()
+            .WithMessage("*collides with the partition key attribute name*");
+    }
+
+    [Fact(Timeout = TestConfiguration.DefaultTimeout)]
+    public void SingleType_WhenDiscriminatorNameCollidesWithSortKey_Throws()
```
ncipollina, line 543: 💡 Missing positive-path test for single-type table

The three new SingleType_ tests all cover failure cases. There is no test confirming that a plain single-type table with the convention-configured discriminator passes validation cleanly. A SingleType_WithConventionDiscriminator_IsValid case would close that gap and serve as the happy-path anchor, mirroring what SharedTableMultipleTypes_WithConventionDiscriminator_IsValid does for the shared-table scenario.

tests/EntityFrameworkCore.DynamoDb.Tests/Metadata/DiscriminatorMetadataTests.cs
```diff
@@ -182,13 +182,14 @@ public static SplitThenSharedTableContext Create(IAmazonDynamoDB client)
     }
 
     [Fact(Timeout = TestConfiguration.DefaultTimeout)]
-    public void SingleTableSingleType_DoesNotConfigureDiscriminatorByConvention()
+    public void SingleTableSingleType_UsesDefaultDiscriminatorByConvention()
```
ncipollina, line 185: ✅ Test rename matches the new default

Renaming from `DoesNotConfigureDiscriminatorByConvention` to `UsesDefaultDiscriminatorByConvention` is exactly right — it now documents the new expected behavior instead of the old absence.

src/EntityFrameworkCore.DynamoDb/Metadata/Conventions/DynamoDiscriminatorConvention.cs
```diff
@@ -145,37 +137,9 @@ private void ProcessModel(IConventionModel model, bool removeSingleTypeConventio
...
-
-                continue;
-            }
+                SetDiscriminatorDisabledAnnotation(rootEntityType, false);
 
-            foreach (var rootEntityType in tableGroup)
-            {
                 var discriminatorBuilder = rootEntityType.Builder.HasDiscriminator(
                     rootEntityType.Model.GetEmbeddedDiscriminatorName(),
                     typeof(string));
```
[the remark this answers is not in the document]
j-d-ha: Addressed in b046b6e by validating discriminator attribute names against all mapped properties, not only PK/SK, and adding single-type collision coverage.

src/EntityFrameworkCore.DynamoDb/Infrastructure/Internal/DynamoModelValidator.cs
j-d-ha: Done in b046b6e. Dropped stale "shared" wording and old multi-type guidance from this guard.

tests/EntityFrameworkCore.DynamoDb.IntegrationTests/SharedTable/DiscriminatorQueryTests.cs
j-d-ha: Done in b046b6e. Renamed to `SingleTypeTable_ProjectsDiscriminator_ButDoesNotInjectWherePredicate`.

tests/EntityFrameworkCore.DynamoDb.Tests/Metadata/DiscriminatorValidationTests.cs
j-d-ha: Done in b046b6e. Added `SingleType_WithConventionDiscriminator_IsValid`.

src/EntityFrameworkCore.DynamoDb/Infrastructure/Internal/DynamoModelValidator.cs
j-d-ha: Done in b046b6e. Added invariant note before loop.

tests/EntityFrameworkCore.DynamoDb.Tests/Metadata/DiscriminatorValidationTests.cs
j-d-ha: Addressed in b046b6e with `SingleType_WithConventionDiscriminator_IsValid` and `SingleTypeValidContext`.

docs/limitations.md
```diff
@@ -394,22 +394,19 @@ update the token value before calling `SaveChangesAsync`.
...
     names.
```
ncipollina: **Docs lag the implementation — mapped-attribute collision is missing from this list**

`ValidateDiscriminatorAttributeNameCollisions` (added in this PR) rejects collisions with *any* mapped property, not only PK/SK. The third bullet here only mentions PK/SK, so users who call `b.Property(e => e.Kind).HasAttributeName("$type")` will hit a startup error that the docs do not describe.

Suggested update to the third bullet:
```
- The discriminator attribute name must not collide with any mapped property attribute name
  (partition key, sort key, or regular properties).
```

This keeps the constraint list complete without requiring a separate bullet for each category.

ncipollina (changes requested):
## Re-review of feedback remediation — 4/5 issues resolved

### Issue verdicts

| # | Issue | Status |
|---|-------|--------|
| 1 | Stale "shared" wording in inner-loop error message | PASS |
| 2 | Misleading test name renamed | PASS |
| 3 | `SingleType_WithConventionDiscriminator_IsValid` added | PASS |
| 4 | `SingleType_WithHasNoDiscriminator_IsValid` added | **FAIL — not present** |
| 5 | Invariant comment restored in inner loop | PASS |

### Issue 4 detail

`SingleType_WithHasNoDiscriminator_IsValid` does not appear anywhere in `DiscriminatorValidationTests.cs`. The file has `SharedTableMultipleTypes_WithHasNoDiscriminator_IsValid` (multi-type case) and `SingleType_WithConventionDiscriminator_IsValid` (positive-path single-type), but no single-type opt-out test. The author's reply to the original comment conflated this with issue 3.

The missing test would cover: a single entity type on its own table calling `HasNoDiscriminator()` — a valid user pattern for existing tables that predate the convention. The validator's early-return guard (`concreteTypes.Count <= 1 || rootEntityTypes.All(IsDiscriminatorDisabled)`) covers this path but is currently untested.

Suggested test skeleton:

```csharp
private sealed class SingleTypeHasNoDiscriminatorContext(DbContextOptions options) : DbContext(options)
{
    protected override void OnModelCreating(ModelBuilder modelBuilder)
        => modelBuilder.Entity<UserEntity>(b =>
        {
            b.ToTable("Users");
            b.HasPartitionKey(x => x.PK);
            b.HasSortKey(x => x.SK);
            b.HasNoDiscriminator();
        });

    public static SingleTypeHasNoDiscriminatorContext Create(IAmazonDynamoDB client)
        => new(BuildOptions<SingleTypeHasNoDiscriminatorContext>(client));
}

[Fact(Timeout = TestConfiguration.DefaultTimeout)]
public void SingleType_WithHasNoDiscriminator_IsValid()
{
    var client = MockClient();
    using var context = SingleTypeHasNoDiscriminatorContext.Create(client);

    var act = () => context.Model;

    act.Should().NotThrow();
}
```

### Sanity check — new finding

One new inline comment was posted on `docs/limitations.md` line 406. The constraints bullet list says the discriminator attribute name must not collide with the partition key or sort key attribute names, but `ValidateDiscriminatorAttributeNameCollisions` now also rejects collisions with any other mapped property. A user who maps a regular property to `$type` will get a startup error not described in the docs. The fix is a one-line update to broaden the third bullet.

### Disposition

**Request Changes** — one item remains unaddressed (issue 4) and one new minor docs inconsistency was found. Both are small fixes; the code itself is correct.

docs/limitations.md
ncipollina: Fixed — SingleTypeHasNoDiscriminatorContext inner class and SingleType_WithHasNoDiscriminator_IsValid test are both present on the branch. Resolving.

ncipollina (dismissed):
All prior feedback addressed. Approving.

PR #267, opened 2026-06-29
test(spec): expand primitive collection query coverage

j-d-ha:
## Summary

Expands `PrimitiveCollectionsQueryDynamoTest` coverage by unskipping additional primitive collection query shapes that now execute successfully, especially EF11-only collection types and direct DynamoDB LIST operations. Also updates provider translation code for primitive collection helper shapes and adds SQL baselines so future query-pipeline changes catch regressions in generated PartiQL, not only result shape. Updates spec coverage docs and the primitive collection audit so remaining skips are categorized around current provider gaps instead of already-fixed quick wins.

### Provider translation

- Extended primitive collection translation in `DynamoSqlTranslatingExpressionVisitor` for supported `Count`/`Length`/`Any` shapes over DynamoDB LIST attributes.
- Added translation for direct first-element access (`First`/`FirstOrDefault`) as LIST index `0` where semantics are provider-supported.
- Added support for `MemoryExtensions.Contains` and list-initializer inline values in finite scalar `Contains` translations.
- Preserved explicit rejection for unsupported primitive collection query pipelines and ambiguous/null-sensitive shapes.

### Spec coverage

- Added `Check_all_tests_overridden` to keep the Dynamo primitive collection spec inventory complete as upstream EF adds or renames inherited test methods.
- Unskipped additional supported parameter/static collection `Contains` shapes under EF11, including `FrozenSet`, `IReadOnlySet`, `ReadOnlyCollectionWithContains`, static `List`, and static `FrozenSet` cases.
- Unskipped supported column collection string `Contains` coverage for EF11 while keeping null-containing shapes skipped because provider primitive collection null semantics remain unsupported there.
- Added explicit skips for still-unsupported shapes (`Parameter_with_inferred_value_converter`, multidimensional arrays, EF11 primitive collection subquery) so current failures are documented and intentional.

### PartiQL baselines

- Added `AssertSql` coverage for representative scalar `IN` translations:
  - inline integer collection `Contains`
  - parameter collection `Contains`
  - `MemoryExtensions.Contains`
- Added baselines for DynamoDB LIST operations:
  - `size("ints") = 2` for `Count`/`Length`
  - `"ints"[1] = 10` for indexer/`ElementAt`
  - `"ints"[0] = 1` for `First`
  - `size("ints") > 0` for `Any`

These baselines make provider intent clearer: simple primitive collection filters should stay server-translated to DynamoDB PartiQL instead of regressing to unsupported/client-only shapes.

### Documentation

- Updated `docs/spec-test-coverage.md` to reflect that primitive collection coverage now differs by target: 43 supported methods on EF10 and 49 on EF11.
- Refreshed `docs/spec-test-primitive-collections-audit.md` to move simple scalar `Contains`, parameter collection `Contains`, `Any`, `size()`, and direct indexing out of the quick-win bucket and into completed coverage.
- Reclassified remaining work around nullable `IN` semantics, value-converted nullable edge cases, inline finite collection aggregate/predicate rewrites, and primitive LIST projection/materialization.
- Clarified long-term blocks for list-element ordering/filtering, owned types, huge in-memory collections, equality/`SequenceEqual`, and list projection shaping.

## Validation

- `task test:spec CONFIG="Debug EF10"` — passed: 2137 total, 1374 succeeded, 763 skipped.
- `task test:spec CONFIG="Debug EF11"` — passed: 2128 total, 1355 succeeded, 773 skipped.

## Notes for Reviewers

- Main review focus: skip/unskip classification in `PrimitiveCollectionsQueryDynamoTest`, provider translation changes in `DynamoSqlTranslatingExpressionVisitor`, and whether newly asserted PartiQL baselines match desired provider behavior.
- This PR records existing supported translations, makes provider translation coverage explicit, and narrows remaining primitive collection gaps.
- EF11-only overrides are guarded with `#if NET11_0_OR_GREATER` because upstream spec surface differs from EF10.


tests/EntityFrameworkCore.DynamoDb.SpecificationTests/Query/PrimitiveCollectionsQueryDynamoTest.cs
```diff
@@ -0,0 +1,899 @@
...
+        "DynamoDB provider does not yet support this primitive collection query shape.";
+
+    protected PrimitiveCollectionsQueryDynamoTest(PrimitiveCollectionsQueryDynamoFixture fixture) :
+        base(fixture)
+        => fixture.ClearSql();
+
+    [ConditionalFact]
+    public virtual void Check_all_tests_overridden()
+        => DynamoTestHelpers.AssertAllTestMethodsOverridden(
+            typeof(PrimitiveCollectionsQueryDynamoTest));
```
ncipollina, line 29: ✅ **Excellent inventory sentinel.** `Check_all_tests_overridden` using `DynamoTestHelpers.AssertAllTestMethodsOverridden` means the provider's test inventory stays current automatically when upstream EF Core adds or renames inherited test methods — no silent drift. This pattern should be used in every future spec test class.

tests/EntityFrameworkCore.DynamoDb.SpecificationTests/Query/PrimitiveCollectionsQueryDynamoTest.cs
```diff
@@ -0,0 +1,899 @@
...
+            """);
+    }
+
+    [ConditionalFact]
+    public override Task Column_collection_index_string() => base.Column_collection_index_string();
+
+    [ConditionalFact]
+    public override Task Column_collection_index_datetime()
+        => base.Column_collection_index_datetime();
+
```
ncipollina, line 886: ⚠️ **Generic skip reason hides DynamoDB-specific semantics.** `Column_collection_index_beyond_end` is skipped with the generic `UnsupportedPrimitiveCollectionQueryShape` constant, but the actual reason is that DynamoDB returns `NULL` for out-of-bounds list index access rather than throwing an exception like the upstream test expects. A more specific skip reason (e.g. `SkipReason.OutOfBoundsListIndexReturnsNullNotThrows` or an inline comment) would make this clearer for future audits and prevent someone from accidentally unskipping it.

tests/EntityFrameworkCore.DynamoDb.SpecificationTests/Query/PrimitiveCollectionsQueryDynamoTest.cs
```diff
@@ -0,0 +1,899 @@
...
+    }
+
+    [Collection(DynamoSpecificationCollection.Name)]
+    public sealed class PrimitiveCollectionsQueryDynamoTestDefault
+        : PrimitiveCollectionsQueryDynamoTest
+    {
+        public PrimitiveCollectionsQueryDynamoTestDefault(
+            PrimitiveCollectionsQueryDynamoFixture fixture,
+            DynamoSpecificationContainerFixture containerFixture) : base(fixture)
+            => _ = containerFixture;
```
ncipollina, line 1252: 💡 **Consider a comment explaining the discard.** `_ = containerFixture` works correctly and is a known pattern here, but a new contributor reading this cold won't know why the fixture is accepted as a parameter but discarded. A brief inline comment like `// ensures DynamoDB Local container is started before tests run` would make the intent immediately clear.

src/EntityFrameworkCore.DynamoDb/Query/Internal/DynamoSqlTranslatingExpressionVisitor.cs
```diff
@@ -1560,18 +1642,64 @@ and not SqlParameterExpression
...
+            return QueryCompilationContext.NotTranslatedExpression;
+
+        SqlExpression? result = null;
+        foreach (var value in values)
+        {
+            var body =
+                new ReplacingExpressionVisitor([lambda.Parameters[0]], [value]).Visit(lambda.Body);
+            if (TranslateInternal(body) is not SqlExpression predicate)
+                return QueryCompilationContext.NotTranslatedExpression;
+
```
ncipollina, line 1680: ✅ **Empty-collection semantics are correct.** When `values` is empty (the inline collection has zero elements), `result` stays `null` and the method returns `Constant(aggregateOperator == AndAlso)`. That gives `false` for empty `Any` and `true` for empty `All` — exactly the standard mathematical identities. This is the right behaviour.

src/EntityFrameworkCore.DynamoDb/Query/Internal/DynamoSqlTranslatingExpressionVisitor.cs
```diff
@@ -1600,8 +1728,37 @@ private static bool TryGetNativePrimitiveCollectionElementMapping(
...
+        SqlExpression collectionExpression,
+        out CoreTypeMapping elementMapping)
+    {
+        elementMapping = null!;
+
+        if (collectionExpression is not SqlPropertyExpression and not DynamoScalarAccessExpression)
+            return false;
+
+        if (!DynamoTypeMappingSource.TryGetListElementType(collectionExpression.Type, out _))
+            return false;
```
ncipollina, line 1748: ✅ **Correct LIST-vs-SET separation.** `TryGetNativePrimitiveListElementMapping` checks `TryGetListElementType` exclusively (unlike the existing `TryGetNativePrimitiveCollectionElementMapping` which accepts both LIST and SET). This correctly prevents `ElementAt`/`First`/`FirstOrDefault` from being applied to SET attributes, which have no positional semantics in DynamoDB.

src/EntityFrameworkCore.DynamoDb/Query/Internal/DynamoSqlTranslatingExpressionVisitor.cs
```diff
@@ -1667,6 +1841,35 @@ private static bool IsCollectionAnyWithoutPredicateMethod(MethodInfo method)
...
+            && (method.GetGenericMethodDefinition() == EnumerableAllMethod
+                || method.GetGenericMethodDefinition() == QueryableAllMethod);
+
+    /// <summary>Returns whether a method represents a primitive-collection Count() call without predicate.</summary>
+    private static bool IsCollectionCountWithoutPredicateMethod(MethodInfo method)
+        => method.IsGenericMethod
+            && (method.GetGenericMethodDefinition() == EnumerableCountMethod
+                || method.GetGenericMethodDefinition() == QueryableCountMethod);
+
+    private static LambdaExpression? UnwrapLambda(Expression? expression)
```
ncipollina, line 1862: ✅ **Both lambda forms handled.** `UnwrapLambda` correctly handles a bare `LambdaExpression` (from `Enumerable` LINQ methods) and a `Quote`-wrapped `LambdaExpression` (from `Queryable` LINQ methods, where the expression tree wraps lambdas in `Quote` nodes). Missing the `Quote` case is a common bug in expression visitors — good catch.

src/EntityFrameworkCore.DynamoDb/Query/Internal/DynamoSqlTranslatingExpressionVisitor.cs
```diff
@@ -1678,6 +1881,15 @@ private static bool IsCollectionContainsMethod(MethodInfo method)
             && method.GetGenericMethodDefinition() == QueryableContainsMethod)
             return true;
 
+        if (method.DeclaringType == typeof(MemoryExtensions)
```
ncipollina, line 1884: 💡 **`MemoryExtensions.Contains` 2-param path could use a type-specific guard.** The current check accepts any `MemoryExtensions.Contains(...)` with 2 bool-returning params. In practice this is fine because the only relevant overloads for a LINQ-over-entities context are `Contains<T>(ReadOnlySpan<T>, T)`. However, adding `method.IsGenericMethod` to the 2-param arm would make the intent explicit and guard against future overloads that happen to fit the same signature. Minor, but would make the guard self-documenting.

src/EntityFrameworkCore.DynamoDb/Query/Internal/SqlExpressionFactory.cs
```diff
@@ -120,11 +120,15 @@ public SqlInExpression In(
     {
         SqlInExpression.ValidateValueSource(null, valuesParameter);
         var mappedItem = ApplyDefaultTypeMapping(item);
-        ThrowIfConvertedEnumComparedToUnderlyingNumber(mappedItem, valuesParameter);
+        var mappedValuesParameter =
+            valuesParameter.TypeMapping is null && mappedItem.TypeMapping is not null
+                ? (SqlParameterExpression)ApplyTypeMapping(valuesParameter, mappedItem.TypeMapping)
+                : valuesParameter;
+        ThrowIfConvertedEnumComparedToUnderlyingNumber(mappedItem, mappedValuesParameter);
```
ncipollina, line 127: ✅ **Correct nullable `IN` mapping fix.** Propagating the mapped item's type mapping to the values parameter when the parameter has no mapping of its own ensures that nullable-collection parameters get the right element converter applied. The guard `valuesParameter.TypeMapping is null && mappedItem.TypeMapping is not null` correctly avoids clobbering an already-mapped parameter. This closes the `List<WrappedId>` vs. `WrappedId` comparison gap.

src/EntityFrameworkCore.DynamoDb/Storage/DynamoTypeMappingSource.cs
```diff
@@ -194,6 +194,12 @@ internal static bool TryGetListElementType(Type clrType, out Type elementType)
     {
         if (clrType.IsArray)
         {
+            if (clrType.GetArrayRank() != 1)
+            {
+                elementType = null!;
```
ncipollina, line 199: ✅ **Clean early-exit for multidimensional arrays.** The `GetArrayRank() != 1` guard is a one-liner that correctly rejects `int[,]`, `string[,,]` etc. before they can enter list-element mapping logic. The matching spec test `Multidimensional_array_is_not_supported` verifies the exception message carries both the property name and the CLR type — good regression coverage.

src/EntityFrameworkCore.DynamoDb/Storage/Internal/DynamoValueReaderWriter.cs
```diff
@@ -163,6 +163,14 @@ internal sealed class DynamoConvertedValueReaderWriter<TModel, TProvider>(
             nameof(Read),
             [typeof(AttributeValue), typeof(string), typeof(bool), typeof(IProperty)])!;
 
+    private static readonly MethodInfo WriteMethod =
+        typeof(DynamoValueReaderWriter<TModel>).GetMethod(nameof(Write), [typeof(TModel)])!;
+
+    private static readonly MethodInfo ToPartiQlLiteralMethod =
```
ncipollina, line 171: ⚠️ **Reflection lookups rely on null-forgiveness (`!`) consistent with existing pattern but worth noting.** `WriteMethod` and `ToPartiQlLiteralMethod` are looked up with `!`, the same as the existing `ReadMethod` and `Constructor` fields above. This is safe as long as the `public abstract Write(TValue)` and `public abstract ToPartiQlLiteral(TValue)` signatures on `DynamoValueReaderWriter<TValue>` remain stable. If those signatures ever change (parameter rename, overload addition), the `!null` would become a silent `NullReferenceException` at class initialization. Consider a small `ThrowIfNull` assertion at class-init time or an explicit type-check, consistent with how you'd guard any critical reflection lookup.

src/EntityFrameworkCore.DynamoDb/Storage/Internal/DynamoValueReaderWriter.cs
```diff
@@ -234,12 +242,31 @@ protected override Expression CreateConstructorExpression()
...
+                ConvertToModelValueExpression(typedValueExpression))
+            : innerReaderWriter.CreatePartiQlLiteralExpression(
+                CreateProviderValueExpression(typedValueExpression));
+
+    private static Expression ConvertToModelValueExpression(Expression typedValueExpression)
+        => typedValueExpression.Type == typeof(TModel)
+            ? typedValueExpression
+            : Expression.Convert(typedValueExpression, typeof(TModel));
+
+    private bool RequiresBoxedConverterFallback(Expression modelValueExpression)
```
ncipollina, line 269: ✅ **Correct heuristic for nullable converter fallback.** Checking whether `ConvertToProviderExpression.Parameters.Single().Type` is a nullable value type tells us the converter was built for a nullable model type (e.g. `CollectionStatus?`). In that case the expression-tree fast path can't safely inline the conversion because `typedValueExpression.Type` may be `object` when pulled from a list index, so routing through the boxed `Write`/`ToPartiQlLiteral` methods is the right fallback. The companion `ConvertToModelValueExpression` that inserts an `Expression.Convert` when needed correctly keeps the tree type-safe.

tests/EntityFrameworkCore.DynamoDb.SpecificationTests/Query/PrimitiveCollectionsQueryDynamoTest.cs
```diff
@@ -0,0 +1,899 @@
...
+    public virtual void Check_all_tests_overridden()
+        => DynamoTestHelpers.AssertAllTestMethodsOverridden(
+            typeof(PrimitiveCollectionsQueryDynamoTest));
+
+    [ConditionalFact]
+    public override async Task Inline_collection_of_ints_Contains()
+    {
+        await base.Inline_collection_of_ints_Contains();
+
+        AssertSql(
```
ncipollina, line 36: ✅ **AssertSql baselines for PartiQL regressions are well-chosen.** Capturing the full SELECT column list along with the WHERE clause means any column-set change, schema rename, or predicate drift will fail the baseline test, not just result-shape differences. This is much stronger than result-only checks and makes the generated PartiQL a first-class contract.

src/EntityFrameworkCore.DynamoDb/Query/Internal/DynamoSqlTranslatingExpressionVisitor.cs
```diff
@@ -1474,33 +1560,29 @@ private Expression TranslateStringStartsWith(MethodCallExpression node)
...
-    ///     Translates <c>Enumerable.ElementAt(source, index)</c> or
-    ///     <c>Queryable.ElementAt(source, index)</c> to a list index access expression. EF Core normalises
-    ///     <c>list[i]</c> to this form, wrapping the source in <c>.AsQueryable()</c> when the collection
-    ///     is a complex collection property.
+    ///     Translates list element access methods to a list index access expression. EF Core normalises
+    ///     <c>list[i]</c> to <c>ElementAt</c>, wrapping the source in <c>.AsQueryable()</c> when the
+    ///     collection is a complex collection property.
     /// </summary>
-    private Expression TranslateElementAt(MethodCallExpression node)
+    private Expression TranslateListIndex(MethodCallExpression node, Expression indexExpression)
```
ncipollina, line 1567: ✅ **Clean unification of list-index translation.** Renaming `TranslateElementAt` to `TranslateListIndex` and accepting the index as a parameter lets `First`/`FirstOrDefault` share the same path as `ElementAt`/indexer with zero duplication. Passing `Expression.Constant(0)` for first-element access is clean and the `ConstantExpression { Value: int index }` pattern match handles it correctly.

ncipollina (dismissed):
## Review Summary

This PR is a substantial, well-executed expansion of primitive collection query coverage for the DynamoDB EF Core provider. The scope spans provider translation, value reader/writer fixes, spec test infrastructure, and documentation — all done with consistent quality. I'm approving; the two flagged items are minor and neither is blocking.

### What this gets right

**Translation correctness**
- `TranslateInlineCollectionPredicate` correctly handles empty-collection identity values (`false` for `Any`, `true` for `All`), uses `ReplacingExpressionVisitor` to substitute lambda parameters before translation, and chains results with the right operator.
- `UnwrapLambda` handles both bare `LambdaExpression` (Enumerable LINQ) and `Quote`-wrapped form (Queryable LINQ) — missing the `Quote` case is a common visitor bug.
- `TryGetNativePrimitiveListElementMapping` correctly scopes `ElementAt`/`First`/`FirstOrDefault` to LIST attributes only (not SET), which have no positional semantics in DynamoDB.
- `TranslateListIndex` unification removes duplication cleanly; passing `Expression.Constant(0)` for `First`/`FirstOrDefault` is the right approach.

**Bug fixes**
- `SqlExpressionFactory.In` propagating the item type mapping to the values parameter when the parameter has no mapping fixes nullable collection `IN` comparisons with value-converted items.
- `TryGetListElementType` multidimensional-array guard (`GetArrayRank() != 1`) is a single, targeted fix that correctly blocks `int[,]` and similar shapes from entering list-element mapping logic.
- `DynamoConvertedValueReaderWriter.RequiresBoxedConverterFallback` correctly identifies nullable-model-type converters that cannot safely inline the conversion expression when the value comes from a list index.
- `MemoryExtensions.Contains` support correctly handles both the 2-arg and 3-arg overloads, with the right short-circuit for a non-null `IEqualityComparer` argument.

**Test quality**
- `Check_all_tests_overridden` is an excellent inventory sentinel — every future spec test class should include this.
- `AssertSql` baselines capture the full SELECT column list plus the WHERE clause, making PartiQL a first-class contract rather than an implicit side-effect.
- Unit tests for `ElementAt`/`First`/`FirstOrDefault` on set and value-converted collections correctly verify rejection paths, not just happy paths.

**Documentation**
- The limitations, operators, and projection docs are updated precisely to match the newly supported and still-unsupported shapes.
- The primitive collections audit doc (`spec-test-primitive-collections-audit.md`) is a useful design record that separates provider gaps from architectural blocks.

### Items worth following up

1. **`Column_collection_index_beyond_end` skip reason** (non-blocking): The skip uses the generic `UnsupportedPrimitiveCollectionQueryShape` constant, but the actual reason is that DynamoDB returns `NULL` for out-of-bounds list index access rather than throwing. A more specific skip message or inline comment would help future auditors distinguish "not implemented yet" from "semantically different by design."

2. **Reflection null-forgiveness in `DynamoConvertedValueReaderWriter`** (non-blocking): `WriteMethod` and `ToPartiQlLiteralMethod` use `!` null-forgiveness, consistent with the existing `ReadMethod` and `Constructor` fields. The pattern is safe while the `DynamoValueReaderWriter<TValue>` abstract method signatures remain stable, but a `Debug.Assert` or comment noting the contract would make the coupling explicit.

Both are nits. The core logic, test coverage, and documentation are correct and production-ready.

ncipollina (dismissed):
✅ Still LGTM

tests/EntityFrameworkCore.DynamoDb.SpecificationTests/Query/PrimitiveCollectionsQueryDynamoTest.cs
j-d-ha: Fixed in fef396e: added a specific skip reason documenting DynamoDB out-of-bounds list index access returns NULL while the upstream test expects an exception.

src/EntityFrameworkCore.DynamoDb/Query/Internal/DynamoSqlTranslatingExpressionVisitor.cs
j-d-ha: Fixed in fef396e: added the explicit method.IsGenericMethod guard for the 2-parameter MemoryExtensions.Contains path.

src/EntityFrameworkCore.DynamoDb/Storage/Internal/DynamoValueReaderWriter.cs
j-d-ha: Fixed in fef396e: added an explicit comment documenting the reflection contract for these lookups.

ncipollina (dismissed):
Thanks for addressing the feedback. Here is the resolution status for each item raised in the previous review.

✅ Issue 1 — `Column_collection_index_beyond_end` skip comment clarity

The `OutOfBoundsListIndexReturnsNull` named constant (`"DynamoDB returns NULL for out-of-bounds list index access; upstream test expects an exception."`) captures both the DynamoDB NULL semantics and the mismatch with the upstream test expectation. The skip reason is now self-documenting at the call site. Fully resolved.

✅ Issue 2 — `WriteMethod`/`ToPartiQlLiteralMethod` reflection null-forgiveness (`!`)

The comment block added above the two static fields reads:
> These reflection lookups are tied to the abstract DynamoValueReaderWriter<TValue> contract. If those signatures change, class initialization should fail rather than build invalid trees.

This documents the intent of the `!` null-forgiveness and the fail-fast contract clearly. Fully resolved.

✅ Issue 3 — `MemoryExtensions.Contains` 2-param `IsGenericMethod` guard

The guard arm now reads `method.GetParameters().Length == 2 && method.IsGenericMethod`, which prevents a hypothetical future non-generic 2-param overload from being incorrectly matched. Fully resolved.

⚠️ Issue 4 — `_ = containerFixture` missing explanatory comment

The `PrimitiveCollectionsQueryDynamoTestDefault` constructor still discards `containerFixture` without a comment. A reader unfamiliar with the xUnit collection-fixture lifetime pattern will not know why the parameter exists. A one-liner would close the gap:

```csharp
// Injected to ensure DynamoDB Local is started by the xUnit collection fixture before any test runs.
=> _ = containerFixture;
```

This was a 💡 suggestion (not a blocking warning), so I am approving, but I would appreciate a follow-up commit to add the comment rather than leaving it as a silent discard.

PR #271, opened 2026-07-02
feat(query): translate binary length and equality

j-d-ha:
## Summary

Adds DynamoDB query translation support and specification coverage for binary `byte[]` operations that map cleanly to PartiQL binary attributes. This moves the binary translation work into a reviewable branch after rolling it back out of `main`.

## Changes

- Translate supported binary query shapes:
  - `byte[].Length` via `size(...)`
  - `byte[].SequenceEqual(...)` via binary equality
  - parameterless `Any()` where supported by EF spec surface
- Add `ByteArrayTranslationsDynamoTest` spec coverage with explicit skips for DynamoDB-incompatible byte element operations.
- Register the new spec test in compliance coverage and update query docs/limitations.
- Add integration coverage for binary length/equality/rejected `Contains` behavior.

## Validation

- `task test:spec:ef10 FILTER="FullyQualifiedName~ByteArrayTranslationsDynamoTest"` — passed, 8 total / 3 succeeded / 5 skipped
- `task test:spec:ef11 FILTER="FullyQualifiedName~ByteArrayTranslationsDynamoTest"` — passed, 9 total / 4 succeeded / 5 skipped

## Notes for Reviewers

This branch intentionally includes the spec-test work that was briefly present on `main`; `main` has been reverted so this PR shows the full intended diff.


src/EntityFrameworkCore.DynamoDb/Query/Internal/DynamoSqlTranslatingExpressionVisitor.cs
```diff
@@ -1638,6 +1656,20 @@ private Expression TranslateListIndex(MethodCallExpression node, Expression inde
...
 
+    /// <summary>Translates byte-array SequenceEqual to binary equality.</summary>
+    private Expression TranslateByteArraySequenceEqual(MethodCallExpression node)
+    {
+        var left = TranslateInternal(StripAsQueryable(node.Arguments[0])!);
+        var right = TranslateInternal(StripAsQueryable(node.Arguments[1])!);
+        if (left == null || right == null)
+            return QueryCompilationContext.NotTranslatedExpression;
+
+        if (!IsBinaryScalarExpression(left) && !IsBinaryScalarExpression(right))
```
ncipollina: ⚠️ This guard is `!IsBinaryScalarExpression(left) && !IsBinaryScalarExpression(right)` — an OR-qualify/AND-reject, so only *one* side needs to look like a Binary scalar to pass. `IsByteArraySequenceEqualMethod` accepts any `IEnumerable<byte>.SequenceEqual(IEnumerable<byte>)`, not just `byte[]`, so a mixed case like `entity.ByteListProperty.SequenceEqual(byteArrayConstant)` (a native `List<byte>` DynamoDB List attribute compared against a `byte[]` Binary constant) would slip through and emit `"byteListAttr" = ?` — comparing incompatible DynamoDB attribute types (`L` vs `B`), which can only ever evaluate false instead of being rejected as untranslatable. Suggest requiring **both** operands to satisfy `IsBinaryScalarExpression` before emitting binary equality. Non-blocking — narrow edge case outside the documented/tested surface.

src/EntityFrameworkCore.DynamoDb/Query/Internal/DynamoSqlTranslatingExpressionVisitor.cs
```diff
@@ -1768,6 +1801,13 @@ private Expression TranslateCollectionSize(Expression? sourceArgument)
     ///     Returns the element mapping when the expression is a natively mapped DynamoDB primitive
     ///     list/set. Scalar value-converted collections intentionally do not qualify.
     /// </summary>
+    private static bool IsBinaryScalarExpression(SqlExpression expression)
```
ncipollina, line 1804: ⚠️ `IsBinaryScalarExpression` only checks `expression.Type == typeof(byte[])` and expression kind — it never checks `expression.TypeMapping?.Converter`. If a `byte[]` property has a value converter to a non-binary store type (e.g. `HasConversion<string>()` for base64 storage), the CLR/expression type is still `byte[]`, so this helper would incorrectly treat it as a native Binary attribute and emit `size(...)`/binary `=` against what's actually a String attribute — silently wrong PartiQL rather than a clear rejection. Other helpers in this file (e.g. `TryGetNativePrimitiveCollectionElementMapping`) check the type mapping/converter to distinguish native vs. converted shapes; this one should follow the same pattern. Suggested fix: also require `expression.TypeMapping is DynamoTypeMapping { Converter: null }` before treating the expression as native Binary. Non-blocking — worth a regression test with a converted `byte[]` property in a follow-up.

src/EntityFrameworkCore.DynamoDb/Query/Internal/DynamoSqlTranslatingExpressionVisitor.cs
```diff
@@ -881,6 +888,9 @@ protected override Expression VisitMethodCall(MethodCallExpression node)
             return QueryCompilationContext.NotTranslatedExpression;
         }
 
+        if (IsByteArraySequenceEqualMethod(node.Method))
```
ncipollina, line 891: ✅ `IsByteArraySequenceEqualMethod`/`TranslateByteArraySequenceEqual` correctly resolve the exact 2-argument `Enumerable.SequenceEqual<byte>` overload and exclude the 3-arg `IEqualityComparer<byte>` overload, which correctly falls through to the generic "not translated" error.

src/EntityFrameworkCore.DynamoDb/Query/Internal/DynamoSqlTranslatingExpressionVisitor.cs
```diff
@@ -1377,6 +1387,14 @@ private bool IsEffectivePartitionKey(IReadOnlyProperty property, IReadOnlyEntity
     /// <inheritdoc />
     protected override Expression VisitUnary(UnaryExpression node)
     {
+        if (node.NodeType == ExpressionType.ArrayLength)
```
ncipollina, line 1390: ✅ `ArrayLength` is correctly gated by `IsBinaryScalarExpression`, so it won't accidentally emit `size(...)` for a non-mapped/non-binary array-typed operand.

tests/EntityFrameworkCore.DynamoDb.IntegrationTests/QueryTypeMappingTable/QueryTypeMappingTests.cs
```diff
@@ -121,6 +121,65 @@ public async Task Between_comparison_returns_expected_items()
...
+        result.Should().Equal("ITEM#2");
+
+        AssertSql(
+            """
+            SELECT "pk"
+            FROM "QueryTypeMappingItems"
+            WHERE "binaryValue" = ?
+            """);
+    }
+
```
ncipollina, line 165: ✅ Meaningful test, not trivial — asserts the specific `InvalidOperationException` message rather than just "throws." Seed data lines up correctly with the length/equality assertions in the tests above this one too.

tests/EntityFrameworkCore.DynamoDb.SpecificationTests/Query/Translations/ByteArrayTranslationsDynamoTest.cs
```diff
@@ -0,0 +1,81 @@
+using EntityFrameworkCore.DynamoDb.SpecificationTests.TestUtilities;
```
ncipollina, line 1: ✅ Follows the documented spec-test convention (`Check_all_tests_overridden`, `ClearSql`, nested `...Default` fixture class, `SkipReason` constants instead of inline literals). The `#if NET11_0_OR_GREATER` guard around `Any()` correctly reflects that upstream EF Core only added an `Any()` case starting with the EF11-targeted spec assembly — verified this explains the 8-vs-9 total test count difference between EF10/EF11 reported in the PR description, reproduced locally (EF10: 3 succeeded / 5 skipped / 8 total).

docs/limitations.md
```diff
@@ -63,6 +63,11 @@ supported for native DynamoDB primitive list/set attributes, but not when `Value
 a scalar string or blob by a property value converter. In that case DynamoDB would evaluate substring
 or binary containment rather than collection membership.
 
+`byte[]` properties map to one DynamoDB Binary attribute. The provider translates binary length,
```
ncipollina, line 66: ✅ Docs are accurate and consistent with the code: correctly states `byte[]` maps to one Binary attribute (not a byte collection), lists exactly the translated operations (`Length`, parameterless `Any()`, `SequenceEqual`) and exactly the rejected ones (`Contains(byte)`, indexing, `First()`).

ncipollina (dismissed):
Reviewed with sr-net-reviewer (senior .NET reviewer persona), plus independently building the branch and running the full unit suite (745 passed / 2 skipped) and the new `ByteArrayTranslationsDynamoTest` spec class (3 succeeded / 5 skipped, matching the PR description exactly). Also traced the parameterization path for binary constants (`BinaryDynamoValueReaderWriter.RequiresParameterForPartiQlLiteral => true`) — confirmed no PartiQL injection risk since byte[] values always go through `AppendParameter`, never string interpolation.

✅ Core translation logic (`size(...)` for Length, binary equality for SequenceEqual, rejection of `Contains(byte)`), type mapping, and doc/test coverage are all correct and verified. Also confirmed `BinaryDynamoValueReaderWriter.RequiresParameterForPartiQlLiteral => true` (pre-existing, unchanged by this PR) — binary values are always parameterized, never string-interpolated, so no PartiQL injection risk. Two narrow edge-case findings left inline — neither affects the documented/tested feature surface, so they're non-blocking follow-ups rather than reasons to hold this PR. Approving.

ncipollina (approved):
Both edge-case fixes confirmed: `TranslateByteArraySequenceEqual` now requires both operands to satisfy `IsBinaryScalarExpression` (line 1667), and `IsBinaryScalarExpression` now checks `TypeMapping is DynamoTypeMapping { Converter: null }` to reject converted byte[] properties (line ~1804). Rebuilt the branch clean, 0 warnings/errors. All feedback resolved. Approving.

PR #272, opened 2026-07-04
test(spec): add DynamoDB scalar type coverage

j-d-ha:
## Summary

Adds DynamoDB specification coverage for scalar type behavior and related query/spec tracking updates. Also updates `docs/spec-test-coverage.md` to separate future-supportable provider work from coverage permanently blocked by DynamoDB/PartiQL limitations.

## Changes

- Adds DynamoDB type spec tests for bool, numeric, string, GUID, temporal, and byte-array types.
- Adds/updates implemented spec inventory entries, compliance registration, and skip reasons.
- Documents provider feature backlog versus hard DynamoDB capability limits for future spec-test planning.

## Validation

- Not run in this session; latest change was docs-only.

## Notes for Reviewers

- Focus review on whether the new backlog/limitations split matches DynamoDB PartiQL capabilities and provider direction.


tests/EntityFrameworkCore.DynamoDb.SpecificationTests/ConcurrencyDetectorDisabledDynamoTest.cs
```diff
@@ -15,31 +15,35 @@ public void Check_all_tests_overridden()
...
 
+    [ConditionalTheory(Skip = SkipReason.SyncQueriesNotSupported)]
     public override Task First(bool async)
-    {
-        if (!async)
-            return Task.CompletedTask;
-
-        return ConcurrencyDetectorTest(async c
+        => ConcurrencyDetectorTest(async c
             => await c.Products.Where(p => p.Id == 1).AsUnsafeFilteredQuery().FirstAsync());
```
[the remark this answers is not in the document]
j-d-ha: Already fixed in later commit by branching on the `async` parameter: sync rows now assert unsupported sync execution, while async rows still run the concurrency detector path.

tests/EntityFrameworkCore.DynamoDb.SpecificationTests/ConcurrencyDetectorEnabledDynamoTest.cs
```diff
@@ -15,31 +15,34 @@ public void Check_all_tests_overridden()
...
 
+    [ConditionalTheory(Skip = SkipReason.SyncQueriesNotSupported)]
     public override Task First(bool async)
-    {
-        if (!async)
-            return Task.CompletedTask;
-
-        return ConcurrencyDetectorTest(async c
+        => ConcurrencyDetectorTest(async c
             => await c.Products.AsUnsafeFilteredQuery().FirstAsync());
```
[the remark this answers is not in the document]
j-d-ha: Already fixed in later commit by branching on the `async` parameter: sync rows now assert unsupported sync execution, while async rows still run the concurrency detector path.

tests/EntityFrameworkCore.DynamoDb.SpecificationTests/Types/DynamoBoolTypeTest.cs
```diff
@@ -0,0 +1,95 @@
...
+    public override async Task Equality_in_query_with_parameter()
+    {
+        await using var context = Fixture.CreateContext();
+
+        var results = await context.Set<TypeEntity<bool>>()
+            .Where(e => e.Value.Equals(Fixture.Value))
+            .ToListAsync();
+        var result = results.Single();
+
+        Assert.Equal(Fixture.Value, result.Value, Fixture.Comparer);
```
ncipollina, line 32: ✅ Not tautological — this actually round-trips through `Where(e => e.Value.Equals(...))`/an `Expression.Lambda`-built predicate against a real DynamoDB table and asserts via `Fixture.Comparer`. Same solid pattern repeated correctly across all 15 new type test files.

tests/EntityFrameworkCore.DynamoDb.SpecificationTests/Types/DynamoBoolTypeTest.cs
```diff
@@ -0,0 +1,95 @@
...
+using EntityFrameworkCore.DynamoDb.Diagnostics;
+using EntityFrameworkCore.DynamoDb.SpecificationTests.TestUtilities;
+using Microsoft.EntityFrameworkCore;
+using Microsoft.EntityFrameworkCore.TestUtilities;
+using Microsoft.EntityFrameworkCore.Types;
+
+namespace EntityFrameworkCore.DynamoDb.SpecificationTests.Types;
+
+[Collection(DynamoSpecificationCollection.Name)]
+public sealed class DynamoBoolTypeTest(
```
ncipollina, line 11: ⚠️ Nit: this file is near character-for-character duplicated across all 15 new `Types/*.cs` files (only the generic type arg, fixture values, and table name differ). Likely unavoidable given the `TypeTestBase<T, TFixture>` pattern requiring its own xUnit class per type — flagging only as a maintenance note, not a blocker. Not worth a source generator for 15 fixed types.

tests/EntityFrameworkCore.DynamoDb.SpecificationTests/Types/DynamoByteArrayTypeTest.cs
```diff
@@ -0,0 +1,100 @@
...
+    [ConditionalFact(Skip = SkipReason.QueryShapeNotSupported)]
+    public override Task Equality_in_query() => base.Equality_in_query();
+#endif
+
+    public class ByteArrayTypeFixture : TypeFixtureBase<byte[]>, IDynamoSpecificationFixture
+    {
+        public override byte[] Value { get; } = [1, 2, 3];
+        public override byte[] OtherValue { get; } = [4, 5, 6, 7];
+
+        public override Func<byte[], byte[], bool> Comparer { get; } = (a, b) => a.SequenceEqual(b);
```
ncipollina, line 69: ✅ Correctly uses `SequenceEqual` for the comparer since `byte[]` needs structural, not reference, equality — other type fixtures can rely on default value-type equality, but this one specifically needed the override. Good attention to detail.

tests/EntityFrameworkCore.DynamoDb.SpecificationTests/SkipReason.cs
```diff
@@ -37,6 +37,9 @@ public static class SkipReason
     public const string ByteArrayContainsElementNotSupported =
         "DynamoDB PartiQL contains() supports set/list membership, but byte[] maps to a single binary attribute rather than a byte collection.";
 
+    public const string TemporalFunctionNotSupported =
```
ncipollina, line 40: ✅ `TemporalFunctionNotSupported`, `SyncQueriesNotSupported`, and `SyncSaveChangesNotSupported` correctly added as shared constants rather than one-off literals — matches the tests/AGENTS.md convention for recurring provider limitations.

tests/EntityFrameworkCore.DynamoDb.SpecificationTests/CustomConvertersDynamoTest.cs
```diff
@@ -432,13 +442,11 @@ public override Task Can_query_custom_type_not_mapped_by_default_equality(bool a
     public override Task Field_on_derived_type_retrieved_via_cast_applies_value_converter()
         => base.Field_on_derived_type_retrieved_via_cast_applies_value_converter();
 
-    // DynamoDB PartiQL does not support joins.
-    [ConditionalFact(Skip = SkipReason.QueryShapeNotSupported)]
+    [ConditionalFact(Skip = SkipReason.JoinsNotSupported)]
```
ncipollina, line 445: ✅ Good cleanup — replaces prior literal-skip drift with the correct shared `SkipReason.JoinsNotSupported` constant. Exactly the kind of consistency fix the test conventions call for.

tests/EntityFrameworkCore.DynamoDb.SpecificationTests/Query/NorthwindWhereQueryDynamoTest.cs
```diff
@@ -31,7 +31,8 @@ public override Task Where_simple(bool async)
             });
 
     [ConditionalTheory(Skip = SkipReason.QueryShapeNotSupported)]
```
ncipollina, line 33: ✅ Not cosmetic churn — this and similar changes throughout the file replace bare `=> Task.CompletedTask;` skip stubs with `=> base.Method(async);`, per tests/AGENTS.md ("keep override wired to base implementation... preserves future compatibility if skip removed"). Verified 0 regressions from this change across the full suite run.

tests/EntityFrameworkCore.DynamoDb.SpecificationTests/ConcurrencyDetectorDisabledDynamoTest.cs
```diff
@@ -15,35 +15,52 @@ public void Check_all_tests_overridden()
...
+    public override Task Any(bool async) => base.Any(async);
 
-    [ConditionalTheory(Skip = "DynamoDB does not support Count queries.")]
-    public override Task Count(bool async) => Task.CompletedTask;
+    [ConditionalTheory(Skip = SkipReason.CountAggregatesNotSupported)]
+    public override Task Count(bool async) => base.Count(async);
 
-    public override Task Find(bool async) => async ? base.Find(async) : Task.CompletedTask;
+    public override Task Find(bool async)
+    {
```
ncipollina, line 25: ✅ Real behavioral upgrade, not just skip-reason relabeling — sync `Find`/`First`/`Single`/`ToList`/`SaveChanges` rows now call `AssertSyncQueryUnsupported`/`AssertSyncSaveChangesUnsupported` and assert the actual `NotSupportedException` and message content, replacing a prior silent no-op. Meaningful correctness improvement.

tests/EntityFrameworkCore.DynamoDb.SpecificationTests/ComplianceDynamoTest.cs
```diff
@@ -12,8 +15,100 @@ namespace EntityFrameworkCore.DynamoDb.SpecificationTests;
...
+                return candidate;
+
+            directory = directory.Parent;
+        }
+
+        throw new DirectoryNotFoundException(
+            "Could not locate tests/EntityFrameworkCore.DynamoDb.SpecificationTests.");
+    }
+
+    private static IEnumerable<string> FindSkippedNoOpOverrides(
```
ncipollina, line 96: ⚠️ `FindSkippedNoOpOverrides` regex-parses C#-like structure (nested `[Skip=...]` + method body) to catch drift back to no-op skip stubs — inherently a bit fragile for edge cases like nested braces or multi-statement bodies. Well mitigated by the dedicated self-test at line 35 validating the regex against a realistic fixture string, and this is test-only tooling (not shipped provider code), so the risk is contained. Coaching note: if this regex needs new cases later, extend the fixture-based self-test first.

tests/EntityFrameworkCore.DynamoDb.SpecificationTests/ComplianceDynamoTest.cs
```diff
@@ -12,8 +15,100 @@ namespace EntityFrameworkCore.DynamoDb.SpecificationTests;
...
+            RegexOptions.Multiline);
+
+        foreach (Match match in pattern.Matches(source))
+        {
+            var relativePath = Path.GetRelativePath(sourceRoot, path);
+            yield return $"{relativePath}: {match.Groups["method"].Value}";
+        }
+    }
+
     protected override IEnumerable<Type> GetBaseTestClasses()
```
ncipollina, line 112: ✅ `GetBaseTestClasses()` additions all correspond 1:1 to new concrete test files added in this PR and correctly reference upstream base types (not concrete Dynamo test types) per convention. Wiring is consistent and I verified all registrations resolve during the full test run.

tests/EntityFrameworkCore.DynamoDb.SpecificationTests/EntityFrameworkServiceCollectionExtensionsDynamoTest.cs
```diff
@@ -0,0 +1,39 @@
...
+using Microsoft.EntityFrameworkCore;
+using Microsoft.EntityFrameworkCore.TestUtilities;
+
+namespace EntityFrameworkCore.DynamoDb.SpecificationTests;
+
+public sealed class EntityFrameworkServiceCollectionExtensionsDynamoTest()
+    : EntityFrameworkServiceCollectionExtensionsTestBase(DynamoTestHelpers.Instance)
+{
+    [ConditionalFact]
+    public void Check_all_tests_overridden()
```
ncipollina, line 13: ⚠️ Nit: this `Check_all_tests_overridden` uses bespoke reflection instead of the shared `DynamoTestHelpers.AssertAllTestMethodsOverridden` helper used everywhere else. The comment above explains why (a non-virtual base fact must be shadowed via `new`), so it's a justified deviation — worth a one-line addition to tests/AGENTS.md's override-handling section documenting this as the exception case for future authors, so the next person doesn't wonder why this file is different.

tests/EntityFrameworkCore.DynamoDb.SpecificationTests/Query/Translations/Temporal/DateTimeTranslationsDynamoTest.cs
```diff
@@ -0,0 +1,120 @@
...
+    public override Task TimeOfDay() => base.TimeOfDay();
+
+    [ConditionalFact(Skip = SkipReason.TemporalFunctionNotSupported)]
+    public override Task subtract_and_TotalDays() => base.subtract_and_TotalDays();
+
+    public override async Task Parse_with_constant()
+    {
+        await base.Parse_with_constant();
+
+        AssertSql(
```
ncipollina, line 66: ✅ Genuinely exercises translation output — asserts real generated PartiQL via `AssertSql(...)` confirming `DateTime` constant/parameter equality lowers to string comparison as documented. Everything else in this file is correctly skipped with `SkipReason.TemporalFunctionNotSupported` since PartiQL has no temporal functions — accurate skip reasoning, matches actual DynamoDB capability.

docs/spec-test-coverage.md
```diff
@@ -273,52 +271,90 @@ Cosmos DB implements all translation categories; MongoDB implements none.
...
+| Key-addressable `ExecuteDelete`            | `DELETE FROM table WHERE key = ...` is supported for a single item                                          | Bulk update/delete specs where predicates can be reduced to item keys                                          |
+| Key-addressable `ExecuteUpdate`            | `UPDATE table SET/REMOVE ... WHERE key = ...` is supported for a single item                                | Bulk update specs for simple scalar/complex-property assignments over known keys                               |
+| Client-expanded batch/transaction writes   | Batch PartiQL supports up to 25 statements; transactions support up to 100 all-read or all-write statements | Broader bulk specs when the provider can first select keys, then issue per-item statements                     |
+| Temporal equality/range translation        | ISO 8601 strings can be compared lexically when stored in normalized form                                   | More `DateTime` coverage and limited `DateOnly`/`DateTimeOffset`/`TimeOnly`/`TimeSpan` equality or range cases |
```
ncipollina, line 333: ⚠️ Worth double-checking that this caveat ("ISO 8601 strings can be compared lexically when stored in normalized form") gets enforced at the type-mapping layer as temporal coverage grows — true lexical ordering breaks if the provider ever stores mixed-precision or non-UTC-offset timestamps. Already caveated correctly here, just flagging as something to keep an eye on, not a doc error.

docs/spec-test-coverage.md
```diff
@@ -273,52 +271,90 @@ Cosmos DB implements all translation categories; MongoDB implements none.
...
+| Non-Query (top-level) | 20 classes / 377 methods |              — |                        — |  25 classes / 1,403 methods |
+| BulkUpdates           |                        — |              — | 5 classes / 136+ methods |        1 class / 33 methods |
+| Northwind Query       |  8 classes / 458 methods |              — |                        — | 14 classes / 1,398+ methods |
+| Other Query           |  6 classes / 337 methods |              — |                        — |  23 classes / 1,814 methods |
+| Associations          |   3 classes / 42 methods |              — |                        — |  13+ classes / 123+ methods |
+| Translations          |  7 classes / 161 methods |              — |                        — |     9 classes / 171 methods |
+
+______________________________________________________________________
+
+## Provider Feature Backlog vs DynamoDB Limits
```
ncipollina, line 319: ✅ Classification checked against known DynamoDB/PartiQL constraints and found accurate — joins/navigation, GROUP BY/aggregates, set operations, offset Skip, global non-key ordering, and temporal/spatial/bitwise functions are correctly placed under "permanently blocked" (real architectural limits, not just current-provider gaps), while "future-supportable" entries all map to real, currently-available PartiQL capabilities the provider hasn't wired up yet. No misclassified rows found.

ncipollina (dismissed):
Reviewed with sr-net-reviewer (senior .NET reviewer persona). Since the author's validation note ("not run in this session; latest change was docs-only") undersells the actual diff — this PR includes 15 new spec test files plus meaningful edits to ComplianceDynamoTest/ConcurrencyDetector*/NorthwindWhere, not just docs — I had the agent independently build and run the full spec suite against real DynamoDB Local rather than trust the description.

Results: **Debug EF10: 2246 total / 1439 passed / 807 skipped / 0 failed. Debug EF11: 2309 total / 1492 passed / 817 skipped / 0 failed.** Unit test project: 745/747 passed (2 pre-existing skips). `dotnet format --verify-no-changes` shows drift, but confirmed identical on `main` — pre-existing, not introduced here.

✅ Code is solid: new type tests exercise real translation/materialization behavior (not tautological), skip reasons correctly use shared `SkipReason` constants, compliance registration is wired correctly, and the docs backlog/limitation split is accurate against actual PartiQL capabilities. Only minor/nit-level coaching comments inline, none blocking. One process note: please avoid describing a diff as "docs-only" when it includes new test code — worth running the suite before future PRs even when the *latest* incremental commit was docs-only. Approving.

tests/EntityFrameworkCore.DynamoDb.SpecificationTests/EntityFrameworkServiceCollectionExtensionsDynamoTest.cs
j-d-ha: Addressed in latest push: added the documented exception to `tests/EntityFrameworkCore.DynamoDb.SpecificationTests/AGENTS.md` for non-virtual upstream facts that must be shadowed with `new` and use local reflection.

tests/EntityFrameworkCore.DynamoDb.SpecificationTests/Types/DynamoBoolTypeTest.cs
j-d-ha: Thanks for validating the pattern across the type coverage. Keeping this as the shared approach for these spec tests.

tests/EntityFrameworkCore.DynamoDb.SpecificationTests/Types/DynamoBoolTypeTest.cs
j-d-ha: Agreed. Duplication is intentional here to keep xUnit discovery and the EF `TypeTestBase<T, TFixture>` shape straightforward; no generator planned for this fixed set.

tests/EntityFrameworkCore.DynamoDb.SpecificationTests/Types/DynamoByteArrayTypeTest.cs
j-d-ha: Thanks. Keeping the structural comparer specifically for the byte-array fixture.

tests/EntityFrameworkCore.DynamoDb.SpecificationTests/SkipReason.cs
j-d-ha: Thanks. Shared skip constants stay the convention for recurring provider limitations.

tests/EntityFrameworkCore.DynamoDb.SpecificationTests/CustomConvertersDynamoTest.cs
j-d-ha: Thanks. This cleanup keeps the spec suite aligned with the shared skip-reason convention.

tests/EntityFrameworkCore.DynamoDb.SpecificationTests/Query/NorthwindWhereQueryDynamoTest.cs
j-d-ha: Thanks. Keeping skipped overrides wired to base remains the intended pattern so future unskips keep coverage instead of hiding drift.

tests/EntityFrameworkCore.DynamoDb.SpecificationTests/ConcurrencyDetectorDisabledDynamoTest.cs
j-d-ha: Thanks. The sync unsupported assertions are intentional coverage so unsupported sync paths fail loudly if behavior changes.

tests/EntityFrameworkCore.DynamoDb.SpecificationTests/ComplianceDynamoTest.cs
j-d-ha: Agreed. If this regex needs to cover more cases later, the fixture-style self-test should be extended first before changing the parser.

tests/EntityFrameworkCore.DynamoDb.SpecificationTests/ComplianceDynamoTest.cs
j-d-ha: Thanks for checking the registrations. Keeping the compliance inventory tied to upstream base types.

tests/EntityFrameworkCore.DynamoDb.SpecificationTests/Query/Translations/Temporal/DateTimeTranslationsDynamoTest.cs
j-d-ha: Thanks. Temporal translation coverage stays limited to real DynamoDB PartiQL behavior; unsupported temporal members remain explicitly skipped.

docs/spec-test-coverage.md
j-d-ha: Agreed. The caveat is intentional, and future temporal comparison work should verify normalized storage at the type-mapping layer before broadening support.

docs/spec-test-coverage.md
j-d-ha: Thanks for validating the classifications against DynamoDB/PartiQL constraints. Keeping the permanent-vs-future-supportable split as documented.

ncipollina (approved):
Confirmed — the AGENTS.md diff adds exactly the documentation we asked for: the reflection-based `Check_all_tests_overridden` exception (for upstream non-virtual tests shadowed via `new`) is now called out explicitly for future authors. All other feedback was nit/coaching-level and non-blocking. No code or test behavior changed since the prior approval, so no re-verification needed. Approving.

PR #273, opened 2026-07-06
refactor(tests): align spec overrides with thin standard

j-d-ha:
## Summary

Aligns DynamoDB specification tests with the thin-override standard. Custom overrides now delegate to base implementations where possible, centralize skip reasoning, and keep provider-specific behavior documented and guarded.

## Changes

- Refactored skipped/custom spec test overrides across data type, converter, non-query, Northwind, and remaining query areas.
- Added centralized skip reasons via `SkipReason` and normalized override guards.
- Added compliance coverage to detect future thin-override drift.
- Documented custom override audit findings and updated spec coverage notes.

## Validation

- `task test:spec CONFIG="Debug EF10" FILTER="FullyQualifiedName~ComplianceDynamoTest"` — passed 6/6.
- `task test:spec CONFIG="Debug EF11" FILTER="FullyQualifiedName~ComplianceDynamoTest"` — passed 6/6.
- `task docs:build` — passed.
- `git diff --check` — passed.
- Branch pushed to `origin/refactor/tests-spec-thin-overrides-all`.

## Notes for Reviewers

Focus review on whether remaining custom overrides capture real DynamoDB provider differences versus accidental spec drift.


tests/EntityFrameworkCore.DynamoDb.SpecificationTests/ComplianceDynamoTest.cs
```diff
@@ -109,6 +301,62 @@ private static IEnumerable<string> FindSkippedNoOpOverrides(
...
+        {
+            if (!customExecutionTokens.Any(token => body.Contains(token, StringComparison.Ordinal)))
+                continue;
+
+            var relativePath = Path.GetRelativePath(sourceRoot, path);
+            yield return $"{relativePath}: {method}";
+        }
+    }
+
+    private static IEnumerable<(string Method, string Body)> FindOverrideBodies(string source)
```
ncipollina, line 380: ⚠️ `FindOverrideBodies` uses a regex (`\{(?:[^{}]|\{[^{}]*\})*\}`) that only tolerates a single level of brace nesting. Any thinned method whose body has two or more levels of nesting (e.g. a `using`/`try` containing an `if` block, or a lambda with its own block) will not match correctly, and the offender will silently disappear from `FindCustomNoBaseOverrides`/`FindCustomExecutionLogicOverrides` output rather than being flagged. This is text-pattern matching, not reflection/IL diffing, so the guard can pass trivially for a real custom override the moment its body nesting gets slightly deeper than what exists today. Worth adding a regression test with 2+ levels of nesting to prove the guard actually catches that case, or switching to a brace-counting scan instead of a bounded regex.

tests/EntityFrameworkCore.DynamoDb.SpecificationTests/ComplianceDynamoTest.cs
```diff
@@ -58,21 +110,121 @@ public override Task Calls_base()
...
+        Assert.Equal(["SampleDynamoTest.cs: Custom_query_then_base"], offenders);
+    }
+
     protected override Assembly TargetAssembly { get; } = typeof(ComplianceDynamoTest).Assembly;
 
-    private static readonly ISet<string> AllowedLegacyNoOpOverrideFiles =
+    private static readonly ISet<string> AllowedCustomNoBaseOverrides =
+        new HashSet<string>(StringComparer.Ordinal);
+
+    private static readonly ISet<string> ThinOverrideCleanupMethodNames =
```
ncipollina, line 200: ⚠️ `ThinOverrideCleanupMethodNames`/`ThinOverrideCleanupFiles` are a fixed allowlist scoped to exactly the methods thinned in this PR. The PR description says this test guards against 'future thin-override drift,' but as written it only prevents regression on these ~50 named methods in these ~18 files — a brand-new custom override added later in a file/method not on this list is invisible to `Spec_tests_do_not_add_unapproved_custom_no_base_overrides_for_thinned_methods` and `Spec_tests_do_not_add_custom_execution_logic_for_thinned_methods`. Consider clarifying the test names/docs to say 'guards this PR's cleanup set' rather than implying general future-drift protection, or better, run the same checks against every override in the whole spec-test tree (like `Spec_tests_do_not_add_skipped_no_op_overrides` already does) so new files are covered automatically.

tests/EntityFrameworkCore.DynamoDb.SpecificationTests/ComplianceDynamoTest.cs
```diff
@@ -109,6 +301,62 @@ private static IEnumerable<string> FindSkippedNoOpOverrides(
...
+            yield return $"{relativePath}: {method}";
+        }
+    }
+
+    private static IEnumerable<string> FindCustomExecutionLogicOverrides(
+        string sourceRoot,
+        string path,
+        string source)
+    {
+        string[] customExecutionTokens =
```
ncipollina, line 355: ⚠️ `customExecutionTokens` is a fixed substring list (`CreateContext(`, `.ToListAsync(`, `.AsAsyncEnumerable(`, `.AllowScan(`, `.AsUnsafeFilteredQuery(`, `ExecuteWithStrategyInTransactionAsync`). A future custom body using different but equally 'custom execution logic' tokens — e.g. `.SingleAsync(`, `.CountAsync(`, `context.Add(`, `Assert.Throws`, `Assert.Equal` on a query result — sails straight through undetected. This gives false confidence that the compliance test 'guards drift' when it really only catches the six specific call shapes that happened to appear in the pre-refactor bodies. Might be safer to flip the heuristic: flag any override body containing more than just `base.Method(...)`/`return Task.CompletedTask`/`await base...`, rather than allowlisting the tokens known-bad today.

tests/EntityFrameworkCore.DynamoDb.SpecificationTests/BuiltInDataTypesDynamoTest.cs
```diff
@@ -82,80 +72,19 @@ public override Task Can_insert_and_read_with_max_length_set()
...
-
-            var entity2 = await QueryByBinaryKey(context, [1, 2, 3, 4]);
-            Assert.Equal("Xx3", entity2.Ex);
-
-            var entity3 = await QueryByBinaryKey(context, [1, 2, 3, 4, 5]);
-            Assert.Equal("Xx1", entity3.Ex);
-        }
-    }
+    [ConditionalFact(Skip = SkipReason.ForeignKeysNotSupported)]
+    public override Task Can_insert_and_read_back_with_binary_key()
```
ncipollina, line 76: 🚫 This test (and `Can_insert_and_read_back_with_string_key` below) previously ran real, passing coverage: insert 3 binary keys of different lengths, read them back individually, update each, and re-verify — genuine round-trip coverage of DynamoDB binary partition keys. It's now unconditionally skipped with `SkipReason.ForeignKeysNotSupported`, but this test has nothing to do with foreign keys (per its own pre-refactor body and the audit doc's description: 'Inserted, queried, and updated binary partition keys'). The skip reason text ('DynamoDB does not support foreign keys') is misleading for why this specific test no longer runs. If the real blocker is that the shared spec model requires an FK-mapped related entity type just to build, that's worth its own accurate reason string (or a comment) rather than reusing `ForeignKeysNotSupported` verbatim. Separately: the audit doc points to `AddAsync_BinaryPartitionKey_ExplicitKey_PersistsAndMaterializes` as replacement coverage, but that's a single explicit-key scenario — it doesn't cover the multi-length-key + update-after-read sequence this spec test exercised. Worth confirming that gap is genuinely accepted and not accidental drift.

tests/EntityFrameworkCore.DynamoDb.SpecificationTests/BuiltInDataTypesDynamoTest.cs
```diff
@@ -82,80 +72,19 @@ public override Task Can_insert_and_read_with_max_length_set()
...
-            var entity = (await context
-                .Set<StringKeyDataType>()
-                .Where(e => e.Id == "Gumball!")
-                .ToListAsync()).Single();
-
-            Assert.Equal("Gumball!", entity.Id);
-        }
-    }
+    [ConditionalFact(Skip = SkipReason.ForeignKeysNotSupported)]
+    public override Task Can_insert_and_read_back_with_string_key()
```
ncipollina, line 86: 🚫 Same concern as `Can_insert_and_read_back_with_binary_key` above: this test previously ran a real string-key insert/read-back round trip and is now skipped under `SkipReason.ForeignKeysNotSupported`, which doesn't describe why a plain string-key test is unsupported. Please double check the reason string is accurate here (vs. e.g. a model-build constraint) so future readers aren't misled about what's actually blocking this test.

tests/EntityFrameworkCore.DynamoDb.SpecificationTests/CustomConvertersDynamoTest.cs
```diff
@@ -68,78 +68,17 @@ public override Task Can_perform_query_with_ansi_strings_test()
...
-
-            var entity2 = await QueryByBinaryKey(context, [1, 2, 3, 4]);
-            Assert.Equal("Xx3", entity2.Ex);
-
-            var entity3 = await QueryByBinaryKey(context, [1, 2, 3, 4, 5]);
-            Assert.Equal("Xx1", entity3.Ex);
-        }
-    }
+    [ConditionalFact(Skip = SkipReason.ForeignKeysNotSupported)]
+    public override Task Can_insert_and_read_back_with_binary_key()
```
ncipollina, line 72: 🚫 Same `ForeignKeysNotSupported` mislabeling as in `BuiltInDataTypesDynamoTest.Can_insert_and_read_back_with_binary_key` — this method (and `Can_insert_and_read_back_with_string_key` below) tests plain byte-array/string partition keys, not foreign keys. Since this exact pattern repeats across at least two files, it's likely intentional (e.g. these test fixtures share a model that requires an FK-mapped type to build at all), but that reasoning isn't stated anywhere near the skip attribute or in `SkipReason`'s doc comment — worth a one-line comment explaining the real blocker so the label doesn't read as copy-paste of the adjacent FK-specific override.

tests/EntityFrameworkCore.DynamoDb.SpecificationTests/ConcurrencyDetectorEnabledDynamoTest.cs
```diff
@@ -32,38 +32,21 @@ public override Task Find(bool async)
...
-        {
-            AssertSyncQueryUnsupported(context => context.Products.AsUnsafeFilteredQuery().First());
-            return Task.CompletedTask;
-        }
-
-        return ConcurrencyDetectorTest(async context
-            => await context.Products.AsUnsafeFilteredQuery().FirstAsync());
-    }
+    [ConditionalTheory(Skip = SkipReason.OrderedResultSetNotSupported)]
+    public override Task First(bool async) => base.First(async);
```
ncipollina, line 36: ⚠️ Before this PR, `First(async: true)` actually executed (`ConcurrencyDetectorTest(... AsUnsafeFilteredQuery().FirstAsync())`) and passed, exercising real concurrency-detector behavior around `First`. Now the whole theory (both `true` and `false`) is skipped via `SkipReason.OrderedResultSetNotSupported`. If the reason this no longer runs for `async: true` is that plain `base.First(async)` (without the `AsUnsafeFilteredQuery`/`AllowScan` opt-in the old override used) throws due to DynamoDB's scan-protection guard, that's a legitimate provider-shape reason — but it's worth confirming explicitly, since skipping the async case throws away previously-working detector coverage rather than just the sync case that was genuinely unsupported. Same question applies to `ToList` below.

tests/EntityFrameworkCore.DynamoDb.SpecificationTests/TestUtilities/DynamoTestHelpers.cs
```diff
@@ -40,6 +40,21 @@ public void NoSyncTest(Action testCode)
         Assert.Fail("Expected DynamoDB sync query failure.");
     }
 
+    /// <summary>Runs an async test and verifies that DynamoDB sync query execution fails.</summary>
+    public async Task NoSyncTest(Func<Task> testCode)
```
ncipollina, line 44: ✅ Clean, minimal addition — a `Func<Task>` overload of `NoSyncTest` alongside the existing `Action`/bool-async overloads, with a clear one-line XML doc. Good example of the 'thin, surgical, well-documented helper' pattern this PR is otherwise asking test overrides to follow.

tests/EntityFrameworkCore.DynamoDb.SpecificationTests/SkipReason.cs
```diff
@@ -9,6 +9,15 @@ public static class SkipReason
     public const string QueryShapeNotSupported =
         "DynamoDB provider does not support this query shape.";
 
+    public const string GuidNewGuidPredicateNotSupported =
+        "DynamoDB provider does not translate Guid.NewGuid() in server-side predicates.";
+
+    public const string CustomTypeEqualityIssue241 =
```
ncipollina, line 18: ✅ Good pattern: centralizing the per-issue skip reasons (`CustomTypeEqualityIssue241`, `ConditionalBoolValueConversionIssue243`) with the issue number baked into the constant name and message, instead of the previous inline `SkipReason.QueryShapeNotSupported + \" #241\"` string concatenation scattered at call sites. Easier to grep for all tests blocked on a given issue now.

docs/spec-test-custom-overrides-audit.md
```diff
@@ -0,0 +1,158 @@
+---
+title: EF Core Specification Custom Override Audit
+description: Coverage matrix for DynamoDB-specific specification-test override bodies and thin-override dispositions.
+---
+
+# EF Core Specification Custom Override Audit
+
+This matrix started as an audit of executable specification-test overrides whose bodies did not
```
ncipollina, line 8: ✅ This audit is a genuinely useful artifact — it separates 'what the pre-refactor body did' from 'current disposition' and honestly calls out gaps as 'Accepted gap' rather than quietly deleting coverage and hoping no one notices. The 'Gap dispositions' table at the bottom is exactly the kind of paper trail a reviewer needs to distinguish real provider-behavior loss from EF-generic behavior that doesn't need DynamoDB-specific coverage. Nice work keeping this reviewable instead of just squashing it into the diff noise.

tests/EntityFrameworkCore.DynamoDb.SpecificationTests/FindDynamoTest.cs
```diff
@@ -21,13 +21,15 @@ public override void Returns_null_for_int_key_not_in_store()
         => NoSyncTest(() => base.Returns_null_for_int_key_not_in_store());
 
     [ConditionalFact(Skip = SkipReason.NullableKeysNotSupported)]
-    public override void Find_nullable_int_key_tracked() { }
+    public override void Find_nullable_int_key_tracked() => base.Find_nullable_int_key_tracked();
```
ncipollina, line 24: ✅ Good cleanup — turning literal no-op `{ }` skipped overrides into base-delegating one-liners is a nice consistency win even though the `[ConditionalFact(Skip = ...)]` means `base` is never actually invoked at runtime; it keeps the file honest about what the override 'would' do and satisfies the no-op-override compliance check uniformly.

ncipollina (changes requested):
## Summary

This is a well-organized refactor with a genuinely useful audit trail (`docs/spec-test-custom-overrides-audit.md`) that honestly documents accepted coverage gaps rather than hiding them, and the `SkipReason` centralization (including issue-specific constants like `CustomTypeEqualityIssue241`/`ConditionalBoolValueConversionIssue243`) is a clear improvement over the old string-concatenation pattern. Most of the ~48 files are mechanical, low-risk thinning of genuinely no-op or EF-generic behavior (change-tracker/tagging/attach semantics) that doesn't need DynamoDB-specific coverage — that's the right call.

However, I'm requesting changes for two categories of issue:

1. **Compliance-guard scope/fragility** (`ComplianceDynamoTest.cs`): the new "thin-override drift" checks are string/regex-based, not reflection/IL diffing. `FindOverrideBodies`'s brace regex only tolerates one level of nesting, and `FindCustomExecutionLogicOverrides`'s token allowlist (`CreateContext(`, `.ToListAsync(`, etc.) only catches the exact call shapes present in today's pre-refactor bodies. Both are also scoped only to the fixed `ThinOverrideCleanupFiles`/`ThinOverrideCleanupMethodNames` list from this PR, not the whole spec-test tree — so despite the PR description's claim of guarding "future thin-override drift," a new custom override added later outside this list is invisible to these checks. This doesn't block what's already been thinned correctly, but it does mean the compliance test provides less protection than advertised.

2. **Mislabeled skip reason + real coverage loss** (`BuiltInDataTypesDynamoTest.cs`, `CustomConvertersDynamoTest.cs`): `Can_insert_and_read_back_with_binary_key` and `Can_insert_and_read_back_with_string_key` previously ran real, passing round-trip coverage (multi-length binary keys, insert/read/update/re-read) and are now unconditionally skipped under `SkipReason.ForeignKeysNotSupported` — a reason that doesn't describe what these tests actually exercise (plain partition keys, no FK involved). Even if this is an accepted/deliberate gap per the audit doc, the skip reason text should say why these are actually blocked, not borrow the adjacent FK-override's reason.

Everything else (Northwind tracking/tagging thinning, `ConcurrencyDetector*` sync-guard preservation, `FindDynamoTest`/`LoggingDynamoTest` no-op cleanup, `DynamoTestHelpers.NoSyncTest` overload) looks correct and consistent with the thin-override goal. Once the skip-reason accuracy is fixed and the compliance test's actual scope is either widened or clarified in its docs/naming, this should be good to merge.

tests/EntityFrameworkCore.DynamoDb.SpecificationTests/ComplianceDynamoTest.cs
j-d-ha: Replaced the bounded brace regex with a brace-counting scan and added a regression covering nested `using`/`if` blocks with query/assert logic. Compliance now catches the nested override body instead of silently dropping it.

tests/EntityFrameworkCore.DynamoDb.SpecificationTests/ComplianceDynamoTest.cs
j-d-ha: Clarified scope in code/test names: these guards protect the finite cleanup inventory from this refactor, not the whole spec tree. The broader policy still comes from `Spec_tests_do_not_add_skipped_no_op_overrides` plus each class' `Check_all_tests_overridden` guard.

tests/EntityFrameworkCore.DynamoDb.SpecificationTests/ComplianceDynamoTest.cs
j-d-ha: Tightened this guard for the cleanup inventory by adding more known execution/assertion shapes (`SingleAsync`, `CountAsync`, `context.Add`, `Assert.*`) and a nested-body regression. Also renamed it to avoid implying a universal future-drift detector.

tests/EntityFrameworkCore.DynamoDb.SpecificationTests/BuiltInDataTypesDynamoTest.cs
j-d-ha: Fixed the misleading skip reason. These plain key tests now use `SkipReason.SharedDataTypesFixtureRequiresForeignKeys`, and the audit doc explicitly calls out the accepted multi-length/update-after-read coverage gap.

tests/EntityFrameworkCore.DynamoDb.SpecificationTests/BuiltInDataTypesDynamoTest.cs
j-d-ha: Fixed with the same dedicated `SharedDataTypesFixtureRequiresForeignKeys` skip reason so this no longer reads like string keys themselves require foreign keys.

tests/EntityFrameworkCore.DynamoDb.SpecificationTests/CustomConvertersDynamoTest.cs
j-d-ha: Fixed the repeated mislabel here too, plus the same `ConvertToProviderTypesDynamoTest` pattern. Audit doc now names the fixture constraint and accepted gap explicitly.

tests/EntityFrameworkCore.DynamoDb.SpecificationTests/ConcurrencyDetectorEnabledDynamoTest.cs
j-d-ha: Confirmed and left as an intentional thin skip. The old override changed the query shape with scan/unsafe-filter opt-ins; the thinned base call hits provider shape/order constraints, and query materialization/First coverage remains in table query suites. SaveChanges detector sync guards are preserved.

tests/EntityFrameworkCore.DynamoDb.SpecificationTests/ComplianceDynamoTest.cs
ncipollina: ✅ Confirmed resolved — the brace regex was replaced with a proper depth-counting scan (`FindBlockBodyEnd`), and the new `Override_body_detection_handles_deeply_nested_custom_logic` test proves a 3-level-deep nested body (`using` > `if` > `Assert`) is now correctly detected as an offender. Verified in commit 6bc2220.

tests/EntityFrameworkCore.DynamoDb.SpecificationTests/ComplianceDynamoTest.cs
ncipollina: ✅ Confirmed resolved — test names now say `_for_cleanup_inventory` and a comment on `ThinOverrideCleanupMethodNames` explicitly states this guard covers this refactor's finite cleanup set, with the broader tree-wide policy delegated to `Spec_tests_do_not_add_skipped_no_op_overrides` and each class's `Check_all_tests_overridden`. That's an honest scope statement rather than an overclaim. Verified in commit 6bc2220.

tests/EntityFrameworkCore.DynamoDb.SpecificationTests/ComplianceDynamoTest.cs
ncipollina: ✅ Substantially mitigated — `customExecutionTokens` now also matches `.SingleAsync(`, `.CountAsync(`, `context.Add(`, `Assert.Equal`, and `Assert.Throws`, closing the specific gaps I called out, backed by the new nested-body regression test. Still technically a finite list rather than an inverted 'anything but base-call' check, but combined with the scope-clarification from the other comment, this is a reasonable, well-tested improvement. Verified in commit 6bc2220.

tests/EntityFrameworkCore.DynamoDb.SpecificationTests/BuiltInDataTypesDynamoTest.cs
ncipollina: ✅ Confirmed resolved — now uses the dedicated `SkipReason.SharedDataTypesFixtureRequiresForeignKeys` constant instead of the misleading `ForeignKeysNotSupported`, and the audit doc explicitly documents the multi-length-key/update-after-read coverage as an accepted spec-suite gap rather than leaving it implicit. Verified in commit 6bc2220.

tests/EntityFrameworkCore.DynamoDb.SpecificationTests/BuiltInDataTypesDynamoTest.cs
ncipollina: ✅ Confirmed resolved — same `SharedDataTypesFixtureRequiresForeignKeys` fix applied here. Verified in commit 6bc2220.

tests/EntityFrameworkCore.DynamoDb.SpecificationTests/CustomConvertersDynamoTest.cs
ncipollina: ✅ Confirmed resolved — same dedicated skip reason applied here and in `ConvertToProviderTypesDynamoTest.cs` (which had the identical pattern), with the audit doc calling out the fixture constraint by name. Verified in commit 6bc2220.

tests/EntityFrameworkCore.DynamoDb.SpecificationTests/ConcurrencyDetectorEnabledDynamoTest.cs
ncipollina: ✅ Resolved via reply confirmation — thanks for confirming this is an intentional provider-shape limitation (thinned base call hits scan/order constraints without the old `AllowScan`/`AsUnsafeFilteredQuery` opt-in) rather than accidentally dropped coverage. One small suggestion for durability: consider folding this explanation into `docs/spec-test-custom-overrides-audit.md`'s Concurrency detector row so it survives outside the PR thread, but not blocking.

ncipollina (approved):
## Re-review after fix commit 6bc2220 ("fix: address spec cleanup review feedback")

All findings from my previous REQUEST_CHANGES review have been resolved. Resolution table:

| Finding | Status | Evidence |
|---|---|---|
| `ComplianceDynamoTest` brace regex only handled 1 level of nesting | ✅ RESOLVED | Replaced with a proper depth-counting scan (`FindBlockBodyEnd`/`FindExpressionBodyEnd`); new `Override_body_detection_handles_deeply_nested_custom_logic` test proves a 3-level-deep nested body is now correctly caught. |
| Compliance guard's "future drift" scope overstated (fixed allowlist only) | ✅ RESOLVED | Tests renamed to `..._for_cleanup_inventory`, plus a code comment clarifying this guards the finite cleanup set from this refactor, while `Spec_tests_do_not_add_skipped_no_op_overrides` + each class's `Check_all_tests_overridden` cover the broader policy. Honest scope statement rather than overclaim. |
| `customExecutionTokens` allowlist had real gaps (`.SingleAsync(`, `Assert.*`, etc. undetected) | ✅ RESOLVED (substantially) | Token list expanded to include `.SingleAsync(`, `.CountAsync(`, `context.Add(`, `Assert.Equal`, `Assert.Throws`, backed by the new nested-body regression. Still a finite list in principle, but now covers the realistic gap I flagged. |
| `BuiltInDataTypesDynamoTest`/`CustomConvertersDynamoTest`/`ConvertToProviderTypesDynamoTest` binary/string key tests mislabeled `SkipReason.ForeignKeysNotSupported` | ✅ RESOLVED | New dedicated `SkipReason.SharedDataTypesFixtureRequiresForeignKeys` constant applied consistently across all three files; audit doc now explicitly documents the multi-length-key/update-after-read coverage as an accepted spec-suite gap instead of leaving it implicit. |
| `ConcurrencyDetectorEnabledDynamoTest.First`/`ToList` skip async coverage that previously ran | ✅ RESOLVED | Author confirmed via review-thread reply this is an intentional thin skip: the thinned base call (without the old `AllowScan`/`AsUnsafeFilteredQuery` opt-in) hits real provider scan/order constraints, and query materialization coverage remains in the integration table-query suites. Suggested (non-blocking) that this explanation also be folded into the audit doc for durability. |

CI is green (`build`, `Debug EF10`, `Debug EF11`, `validate` all passing). Approving.

PR #275, opened 2026-07-08
ci(github): add dependabot configuration

j-d-ha:
## Summary

Adds Dependabot configuration for this repository, based on the Dynamo Mapper setup. Covers NuGet and non-NuGet dependencies while keeping generated PR volume lower through grouped updates.

This also adds a repo-local required-build adapter so the existing org-managed `build / build` branch protection check reflects this repo's EF10/EF11 application build matrix.

## Changes

- Adds weekly Dependabot updates for `nuget`, `dotnet-sdk`, `github-actions`, `docker-compose`, and `uv` ecosystems.
- Deliberately removes `.opencode` / `npm` from Dependabot coverage because it is local development tooling, not build/release infrastructure.
- Splits NuGet updates into `dotnet-minor-patch` and `dotnet-major` groups, accepting that EF10/EF11 dependency lines may share a grouped PR to reduce update noise.
- Combines `dotnet-sdk`, `github-actions`, `docker-compose`, and `uv` into one `infrastructure` multi-ecosystem group.
- Sets `cooldown.default-days: 7` for each ecosystem so version updates wait one week before PR creation.
- Uses conventional commit formatting with `chore` prefix and dependency scope inclusion.
- Renames the PR build matrix to `application build (...) / build` and adds `build / build` as an aggregate branch-protection adapter.

## Validation

- Parsed `.github/dependabot.yml` with Ruby YAML loader.
- Verified configured ecosystems, multi-ecosystem group assignment, and required `cooldown` / `commit-message` settings.
- Temporarily broke a test to confirm the aggregate `build / build` check fails when an application build matrix leg fails, then restored the test.

## Notes for Reviewers

- GitHub documentation states Dependabot `cooldown` applies to version updates, not security updates.
- Cleaner long-term option: add a Safe Settings repo-level override in `LayeredCraft/.github` so this repo can require `application build (Debug EF10, 10.0.x) / build` and `application build (Debug EF11, 11.0.x) / build` directly. I confirmed direct branch-protection edits are reverted by Safe Settings today.


.github/workflows/pr-build.yaml
```diff
@@ -28,3 +28,11 @@ jobs:
       dotnetVersion: ${{ matrix.dotnetVersion }}
       runCdk: false
     secrets: inherit
+
+  build:
+    name: build
+    needs: application-build
+    if: ${{ always() }}
```
ncipollina, line 35: ✅ Nice touch forcing `if: always()` here. Without it, if `application-build` fails, this wrapper job would be skipped by default (since `needs` implies `success()`), and a *skipped* required check can be treated as passing by GitHub branch protection — silently defeating the whole point of the aggregate check. Forcing this job to always run and then explicitly fail via the reusable workflow closes that gap.

.github/workflows/pr-build.yaml
```diff
@@ -28,3 +28,11 @@ jobs:
       dotnetVersion: ${{ matrix.dotnetVersion }}
       runCdk: false
     secrets: inherit
+
+  build:
+    name: build
+    needs: application-build
```
ncipollina, line 34: ✅ Good future-proofing: this reads `needs.application-build.result`, which GitHub automatically rolls up across the entire matrix (fails if *any* leg fails/cancels). If a third EF version leg is added to the matrix later, no change is needed here or in `required-build-check.yaml` — the aggregation is leg-count-agnostic.

.github/workflows/required-build-check.yaml
```diff
@@ -0,0 +1,26 @@
...
+    name: build
+    runs-on: ubuntu-latest
+    steps:
+      - name: Verify application build
+        run: |
+          case "${{ inputs.applicationBuildResult }}" in
+            success)
+              echo "Application build matrix passed."
+              ;;
+            *)
```
ncipollina, line 22: 👍 Confirmed the case-statement logic is correct: only an exact `success` passes, everything else (failure/cancelled/skipped) falls to the `*` branch and fails the job. That's the right default — a stray `skipped` result won't accidentally satisfy the required check.

.github/dependabot.yml
```diff
@@ -0,0 +1,68 @@
...
+      interval: "weekly"
+      day: "wednesday"
+    open-pull-requests-limit: 25
+    cooldown:
+      default-days: 7
+    commit-message:
+      prefix: "chore"
+      include: "scope"
+    groups:
+      dotnet-minor-patch:
```
ncipollina, line 26: ⚠️ Splitting nuget updates into `dotnet-minor-patch` vs `dotnet-major` groups by update-type (with `patterns: "*"`) means updates across *both* the EF10 and EF11 dependency lines can land in the same PR (e.g. a minor bump to an EF10-targeted package and an unrelated minor bump to an EF11-targeted package get grouped together). If one package in the group has a bad release, it can hold up the whole batch's CI until resolved or excluded. Given this repo dual-targets EF10/EF11 explicitly (see `docs/multi-version-ef-strategy.md`), consider whether splitting by dependency line (or keeping EF Core packages ungrouped) would make it easier to know which PR is safe to merge when one leg's build fails. Not blocking — just flagging the trade-off since grouping strategy was explicitly called out for review.

.github/dependabot.yml
```diff
@@ -0,0 +1,68 @@
...
+      include: "scope"
+
+updates:
+  - package-ecosystem: "nuget"
+    directory: "/"
+    schedule:
+      interval: "weekly"
+      day: "wednesday"
+    open-pull-requests-limit: 25
+    cooldown:
```
ncipollina, line 20: ✅ Correctly scoped — per GitHub docs, `cooldown` only delays *version* updates, not *security* updates, so this won't slow down vulnerability patches. Matches the PR description's own note about cooldown semantics.

.github/dependabot.yml
```diff
@@ -0,0 +1,68 @@
+version: 2
```
ncipollina, line 1: 👍 Checked the `multi-ecosystem-groups` schema (nuget/dotnet-sdk/github-actions/docker-compose/uv, cooldown, grouping) against GitHub's docs — structure looks valid (`multi-ecosystem-groups` at top level with `schedule`/`open-pull-requests-limit`/`commit-message`, referenced per-update via `multi-ecosystem-group`). Commit history ("fix dependabot multi-ecosystem config", "tune dependabot grouping") also shows this was iterated on and presumably validated by GitHub actually parsing it, which is reassuring since there's no great local-lint story for this file.

.github/dependabot.yml
```diff
@@ -0,0 +1,68 @@
...
+    cooldown:
+      default-days: 7
+
+  - package-ecosystem: "uv"
+    directory: "/"
+    patterns:
+      - "*"
+    multi-ecosystem-group: "infrastructure"
+    cooldown:
+      default-days: 7
```
ncipollina, line 68: 🚫 The old npm ecosystem entry for `/.opencode` was dropped, but `.opencode/package.json` still exists on disk with a real dependency (`@opencode-ai/plugin`). Excluding local dev tooling from version-update noise is a reasonable call, but worth a one-line note in the PR description confirming that's deliberate (vs. an accidental drop during the grouping rework), so a future reader doesn't wonder why it disappeared.

ncipollina (approved):
Reviewed as CI/repo-config-only (no application code). Summary of what was verified:

**Dependabot config** — schema for `multi-ecosystem-groups` / `multi-ecosystem-group` and `cooldown` checks out against GitHub's docs; cooldown correctly only affects version updates, not security updates (matches the PR description). Grouping by update-type (minor/patch vs major) is reasonable, though flagged one trade-off inline: grouping all nuget packages together with `patterns: "*"` can mix EF10-line and EF11-line updates into a single PR, so a bad release in one line could hold up the whole batch. Not a blocker, just worth keeping in mind given this repo's dual EF10/EF11 targeting. Also flagged that `.opencode/package.json` (npm) still exists on disk but is no longer covered — likely an intentional scope decision, worth a one-line confirmation in the PR description.

**Workflow rename (`pr-build.yaml`)** — grepped the whole repo for the old `build` job name; the other `needs: build` references (`docs.yaml`, `publish-release.yaml`, `publish-preview.yaml`) are all local to their own workflow files and unaffected by this rename. No dangling references found.

**Aggregate required-check workflow** — traced the logic carefully since this is the crux of the PR: `application-build`'s matrix result rolls up naturally via `needs.application-build.result` (GitHub aggregates matrix legs automatically), the wrapper job correctly uses `if: always()` so it isn't skipped when the matrix fails (avoiding the classic pitfall where a skipped required check gets treated as passing), and `required-build-check.yaml`'s case statement only treats an exact `success` as passing. This design is also leg-count-agnostic — adding a third EF version to the matrix later needs no changes here, which addresses the maintainability concern about the workaround falling out of sync.

Given this is understood by the author as a stopgap workaround (with a cleaner long-term fix already identified via Safe Settings override) and the mechanics check out correctly, approving. The one non-blocking suggestion (nuget grouping trade-off) can be addressed as a fast-follow if desired.

.github/dependabot.yml
j-d-ha: Keeping current NuGet grouping for this PR. Trade-off is intentional: lower Dependabot PR volume wins for now, and a bad EF10/EF11 package can be split/excluded in a follow-up if CI shows pain. Updated PR description to make that explicit.

.github/dependabot.yml
j-d-ha: Confirmed deliberate. `.opencode` is local dev tooling, not build/release infrastructure, so I kept it out of Dependabot coverage and updated the PR description with that rationale.

PR #280, opened 2026-07-09
fix: complete codebase cleanup release prep

j-d-ha:
## Summary

Cleanup and release-prep batch for the DynamoDB EF Core provider. This tightens CI/release workflows, fixes several query and execution edge cases, improves packaging metadata/docs, and finishes repository hygiene tasks tracked in beads.

### CI, release, and packaging

- Hardened docs and PR workflows with narrower permissions, correct `zensical.toml` usage, no inherited PR secrets, and a pinned reusable workflow SHA.
- Replaced external release-drafter template usage with local EF10/EF11 release-drafter jobs.
- Improved NuGet metadata and re-enabled package analysis.

### Query and execution behavior

- Fixed case-sensitive projection alias deduplication.
- Restricted `string.Compare` / `CompareTo` translation to comparisons against `0`.
- Rejected unsupported char string overload translations.
- Normalized reversed inclusive range predicates into `BETWEEN`.
- Added shared 8KB PartiQL statement validation for reads and writes.
- Snapshotted PartiQL read parameters at enumerable creation.
- Disposed provider-owned DynamoDB clients while preserving user-supplied clients.

### Tests, docs, and cleanup

- Added unit/integration coverage for the query, execution, disposal, scalar equality, and statement-size paths.
- Centralized and pinned the DynamoDB Local image.
- Updated querying docs and nav, fixed example property naming, and removed obsolete skipped/no-op test paths.
- Applied formatting and removed boilerplate test XML comments.
- Added cleanup analysis and repo-local skills used during the cleanup pass.

## Validation

- `task test:ef10`
- `task test:ef11`
- `uv run zensical build -f zensical.toml --clean`
- `dotnet format EntityFrameworkCore.DynamoDb.slnx --no-restore --verify-no-changes --verbosity minimal`
- `task build:ef10`
- Focused query/storage/integration tests during implementation, including `DynamoClientWrapperTests`

## Release Notes

- Read and write PartiQL statements now fail deterministically before AWS when statement text exceeds DynamoDB's 8192-byte limit.
- `string.Compare` / `CompareTo` translations now only support comparisons against `0`, matching .NET's negative/zero/positive contract.
- Query parameter values for PartiQL reads are captured when the enumerable is created, not after later caller mutation.

## Notes for Reviewers

This PR is broad but grouped into small commits by concern. Behavior-sensitive areas to review first: query translation changes, `DynamoClientWrapper` parameter snapshot/disposal semantics, and CI/release workflow hardening.


.github/workflows/pr-build.yaml
```diff
@@ -19,15 +20,14 @@ jobs:
             dotnetVersion: "10.0.x"
           - configuration: "Debug EF11"
             dotnetVersion: "11.0.x"
-    uses: LayeredCraft/devops-templates/.github/workflows/pr-build.yaml@v10.1
+    uses: LayeredCraft/devops-templates/.github/workflows/pr-build.yaml@76a2269c95c0f17eaac80d3020c783ff10be4371
```
ncipollina, line 26: ❓ Why did you change this to a commit hash?

ncipollina (changes requested):
Left a comment.

src/EntityFrameworkCore.DynamoDb/Storage/DynamoClientWrapper.cs
```diff
@@ -47,19 +49,39 @@ public virtual IAmazonDynamoDB Client
...
 
+    /// <summary>Disposes the provider-created DynamoDB client, if this wrapper owns it.</summary>
+    public void Dispose()
+    {
+        if (_disposed)
+            return;
+
+        _disposed = true;
+        if (_ownsClient)
+            Client.Dispose();
```
ncipollina, line 70: ⚠️ The dispose logic itself looks correct (only disposes when `_ownsClient` is true, and `Client` is already non-null in that branch so the getter will not lazily re-create a client during teardown). However, I could not find a test that actually exercises the "provider owns the client" path — `DynamoClientWrapperDisposeTests` only covers the negative case (`Disposing_context_does_not_dispose_user_supplied_client`). Since `AmazonDynamoDBClient` is sealed, a direct mock is hard, but a test asserting `_disposed`/`_ownsClient` via reflection, or one that disposes the wrapper twice and confirms no exception (double-dispose safety), would close the gap on the one behavior this PR explicitly calls out as risk-sensitive.

src/EntityFrameworkCore.DynamoDb/Query/Internal/DynamoSqlTranslatingExpressionVisitor.cs
```diff
@@ -865,7 +865,6 @@ protected override Expression VisitMethodCall(MethodCallExpression node)
             return TranslateStringIsNullOrEmpty(node);
 
         if (node.Method == StringContainsMethod
```
ncipollina, line 861: 🎨 Now that `StringContainsCharMethod` (declared line 97) and `StringStartsWithCharMethod` (declared line 108) are no longer matched here, those two `MethodInfo` fields are dead code — nothing else in the file references them. Worth deleting them in this same cleanup pass; leaving them around risks someone reintroducing the char-overload match without realizing the generic `DeclaringType == typeof(string)` fallback below already rejects those overloads correctly.

src/EntityFrameworkCore.DynamoDb/Storage/DynamoPartiQlStatementValidator.cs
```diff
@@ -0,0 +1,32 @@
...
+    {
+        if (!ContainsNonAscii(statement))
+        {
+            if (statement.Length <= MaxStatementLength)
+                return;
+
+            throw new InvalidOperationException(
+                $"The generated PartiQL {operation} statement is {statement.Length} characters "
+                + $"(ASCII-equivalent bytes), which exceeds DynamoDB's "
+                + $"{MaxStatementLength}-byte statement-size limit.");
```
ncipollina: 💡 The previous `DynamoSaveChangesPlanner.ValidateStatementLength` messages included actionable guidance ("Consider reducing the number of mapped scalar properties or splitting the write unit across multiple SaveChanges calls."). That text was dropped when the logic moved into this shared validator — worth restoring per-operation guidance (write vs. read wording differs) so the exception message stays actionable for both callers, not just a size report.

.github/workflows/release-drafter.yaml
```diff
@@ -9,12 +9,42 @@ on:
...
+      matrix:
+        include:
+          - config-name: release-drafter-ef10.yml
+          - config-name: release-drafter-ef11.yml
+    concurrency:
+      group: release-drafter-${{ matrix.config-name }}-${{ github.ref }}
+      cancel-in-progress: true
+    steps:
+      - name: Draft ${{ matrix.config-name }} release
+        uses: release-drafter/release-drafter@v7
```
ncipollina: ⚠️ This PR pins the `LayeredCraft/devops-templates` reusable workflows/actions to a commit SHA elsewhere (e.g. `pr-build.yaml`, `publish-preview.yaml`), but `release-drafter/release-drafter@v7` and `release-drafter/release-drafter/autolabeler@v7` (line 24) are still referenced by a mutable tag. The `draft` job here also runs with `contents: write`, so a tag re-point (accidental or malicious) would execute with write access to the repo. For consistency with the rest of this hardening pass, consider pinning these to a commit SHA too.

.github/workflows/publish-preview.yaml
```diff
@@ -33,15 +35,14 @@ jobs:
             dotnetVersion: "11.0.x"
             drafterConfig: "release-drafter-ef11.yml"
             artifact_name: "nuget-packages-ef11"
-    uses: LayeredCraft/devops-templates/.github/workflows/publish-preview.yml@v10.1
+    uses: LayeredCraft/devops-templates/.github/workflows/publish-preview.yml@76a2269c95c0f17eaac80d3020c783ff10be4371
```
ncipollina, line 38: ⚠️ `secrets: inherit` was removed from this reusable-workflow call as part of the "no inherited PR secrets" hardening. That rationale fits PR-triggered workflows, but this workflow only runs on `push`/`workflow_dispatch` (trusted context, not attacker-controlled), so if the downstream `publish-preview.yml` template needs any secrets during `build` (e.g. private feed auth, signing, codecov) they will now silently be unset rather than failing loudly at the call site. Worth confirming the downstream template does not expect inherited secrets in the `build` job before merging, since a break here would only surface on the next push to `main`.

src/EntityFrameworkCore.DynamoDb/Query/Internal/DynamoQueryableMethodTranslatingExpressionVisitor.cs
```diff
@@ -1147,45 +1147,58 @@ private static bool TryExtractBetweenBounds(
             || andExpression.Right is not SqlBinaryExpression rightBinary)
```
ncipollina, line 1147: ✅ Nice fix: `TryGetBetweenBound` now accepts the property on either side of each comparison (e.g. `low <= item.Prop && item.Prop <= high`), where the old implementation required the property strictly on the left in both clauses. Good coverage too — `BetweenTests.cs` explicitly exercises `LowerBoundOnLeft` and `UpperBoundOnLeft` cases plus the exclusive-bound non-rewrite cases.

src/EntityFrameworkCore.DynamoDb/Query/Internal/Expressions/SelectExpression.cs
```diff
@@ -379,7 +379,7 @@ private int AddProjectionIfNotExists(SqlExpression sqlExpression, string alias)
         // Prefer alias deduplication first so independently-built equivalent SQL expressions still
         // resolve to a stable projection ordinal for the same logical column.
         for (var i = 0; i < _projection.Count; i++)
-            if (string.Equals(_projection[i].Alias, alias, StringComparison.OrdinalIgnoreCase))
+            if (string.Equals(_projection[i].Alias, alias, StringComparison.Ordinal))
```
ncipollina, line 382: ✅ Good catch — `OrdinalIgnoreCase` alias dedup meant `Select(x => new { Name = x.Name, name = x.OtherProp })`-style case-only-distinct projections would silently collapse to one column. Switching to `Ordinal` fixes it, and `AddToProjection_CaseOnlyAliases_AreDistinct` pins the behavior with an explicit regression test.

src/EntityFrameworkCore.DynamoDb/Storage/DynamoClientWrapper.cs
```diff
@@ -47,19 +49,39 @@ public virtual IAmazonDynamoDB Client
...
+
+        GC.SuppressFinalize(this);
+    }
+
     /// <summary>Creates a reusable async enumerable over PartiQL result pages.</summary>
     public IAsyncEnumerable<Dictionary<string, AttributeValue>> ExecutePartiQl(
         ExecuteStatementRequest statementRequest,
         bool singlePageOnly = false,
         Action<ExecuteStatementResponse>? onPageFetched = null,
```
ncipollina, line 79: ✅ Cloning parameters (`cloneParameters: true`) at `ExecutePartiQl` call time correctly fixes the caller-mutation hazard: if the caller reused/mutated the same `List<AttributeValue>` after building the request but before the async enumerable was actually iterated, the old code (`cloneParameters: false`) would send whatever the list looked like at enumeration time, not creation time. `ExecutePartiQl_SnapshotsParametersAtEnumerableCreation` verifies this precisely (mutates original list + replaces `requestPrototype.Parameters` after creation, then asserts the original snapshot was sent).

ncipollina (dismissed):
Reviewed the diff end-to-end, focusing on the areas called out as behavior-sensitive: query translation (string.Compare/CompareTo restriction, BETWEEN normalization, char-overload rejection, projection alias dedup), `DynamoClientWrapper` parameter-snapshot/disposal semantics, and CI/release workflow hardening.

**Correctness:** No bugs found. The `DynamoClientWrapper.Dispose()` logic is sound — it only disposes a provider-owned client, and since `_ownsClient` implies the backing field is already set, the getter won't lazily re-create a client during teardown. The BETWEEN normalization fix (property on either side of the comparison) and the projection alias case-sensitivity fix are both correct and well covered by new tests (`BetweenTests.cs`, `AddToProjection_CaseOnlyAliases_AreDistinct`). The parameter-snapshot fix for PartiQL reads is verified by a precise mutation test.

**Non-blocking findings posted inline:**
- 🧪 No test exercises the "provider owns the client" dispose path (only the user-supplied-client-is-preserved path is tested).
- 🎨 Two now-unused `MethodInfo` fields (`StringContainsCharMethod`, `StringStartsWithCharMethod`) left behind after the char-overload rejection change.
- 💡 The shared `DynamoPartiQlStatementValidator` dropped the actionable remediation text from the old write-path error message.
- ⚠️ `release-drafter@v7` / `autolabeler@v7` are referenced by mutable tag rather than pinned SHA, inconsistent with the rest of this PR's supply-chain hardening, and the `draft` job runs with `contents: write`.
- ⚠️ `secrets: inherit` was dropped from the `publish-preview.yaml` reusable-workflow call; worth confirming the downstream template doesn't need any inherited secrets in that job, since a break would only surface on the next push to `main`.

Also called out several things done well inline (BETWEEN fix + tests, alias dedup fix + regression test, parameter snapshot fix + test). None of the above rise to a correctness or resource-leak bug, so approving.

.github/workflows/pr-build.yaml
j-d-ha: Pinned reusable workflow refs to commit SHAs so workflow code cannot move under a mutable tag/branch. That is why this changed to a commit hash.

src/EntityFrameworkCore.DynamoDb/Storage/DynamoClientWrapper.cs
j-d-ha: Added coverage for the provider-owned client path: the test forces lazy client creation, verifies the wrapper owns it, disposes twice, and asserts the disposed state. Full EF10/EF11 test suites pass.

src/EntityFrameworkCore.DynamoDb/Query/Internal/DynamoSqlTranslatingExpressionVisitor.cs
j-d-ha: Removed the now-unused `StringContainsCharMethod` and `StringStartsWithCharMethod` fields.

src/EntityFrameworkCore.DynamoDb/Storage/DynamoPartiQlStatementValidator.cs
j-d-ha: Restored actionable statement-size guidance. Write errors now keep the SaveChanges scalar/split wording, and read errors include projection/predicate/smaller-request guidance. Added assertions for both paths.

.github/workflows/release-drafter.yaml
j-d-ha: Pinned both `release-drafter/release-drafter` and `release-drafter/release-drafter/autolabeler` to the peeled `v7` commit SHA (`4d75298e00d9e34c483e5ff8c68d0ea1c1940c1e`).

.github/workflows/publish-preview.yaml
j-d-ha: Confirmed the pinned downstream `publish-preview.yml` build job only uses `secrets.GITHUB_TOKEN` for Release Drafter dry-run plus normal restore/build/test/pack/artifact steps. No inherited external secrets are required there, so keeping `secrets: inherit` removed is intentional.

.github/workflows/pr-build.yaml
ncipollina: ✅ Makes sense — pinning to a commit SHA (`76a2269c95c0f17eaac80d3020c783ff10be4371`) prevents workflow code from moving under a mutable tag/branch. Confirmed the same SHA is used consistently across pr-build.yaml, publish-preview.yaml, and publish-release.yaml at c5f24aa.

src/EntityFrameworkCore.DynamoDb/Storage/DynamoClientWrapper.cs
ncipollina: ✅ Confirmed resolved. `Disposing_provider_owned_client_marks_wrapper_disposed` now forces lazy client creation via `DynamoDbClientConfig`, asserts `_ownsClient` is true via reflection, disposes twice (double-dispose safety), and asserts `_disposed` is true afterward. This closes the coverage gap on the provider-owns-client path.

src/EntityFrameworkCore.DynamoDb/Query/Internal/DynamoSqlTranslatingExpressionVisitor.cs
ncipollina: ✅ Confirmed — `StringContainsCharMethod` and `StringStartsWithCharMethod` fields are removed at c5f24aa; verified no remaining references in the file.

src/EntityFrameworkCore.DynamoDb/Storage/DynamoPartiQlStatementValidator.cs
ncipollina: ✅ Confirmed resolved. `GetRemediation(operation)` now restores per-operation guidance (write: "reducing scalar properties / splitting SaveChanges"; read: "narrowing projection / simplifying predicate / smaller requests"), and both paths are pinned by tests (`SaveChangesEdgeCasesTests`, `ReturnConsumedCapacityRequestTests`).

.github/workflows/release-drafter.yaml
ncipollina: ✅ Confirmed resolved. Both `release-drafter/release-drafter@v7` and `release-drafter/release-drafter/autolabeler@v7` are now pinned to the peeled SHA `4d75298e00d9e34c483e5ff8c68d0ea1c1940c1e`. Also noted the `draft` job's top-level permissions were tightened (workflow-level now `contents: read, pull-requests: read`, with `contents: write` only at the job level that needs it), which further reduces blast radius versus the original concern.

.github/workflows/publish-preview.yaml
ncipollina: ⚠️→✅ Author confirmed the pinned downstream `publish-preview.yml` build job only needs `secrets.GITHUB_TOKEN` (auto-injected, not part of `secrets: inherit`) for the Release Drafter dry-run plus standard restore/build/test/pack/artifact steps — no external secrets required. Given the workflow is push/workflow_dispatch-only (trusted context) and CI is green end-to-end (including the actual publish-preview run), I'm satisfied this is resolved. Worth a one-line code comment near the reusable-workflow call noting 'no external secrets needed' so a future contributor doesn't reflexively re-add `secrets: inherit` — optional, not blocking.

ncipollina (approved):
Re-reviewed at c5f24aa (5 new commits since my prior CHANGES_REQUESTED). All 7 items from the last round are now addressed:

- ✅ `DynamoClientWrapper` dispose gap: new `Disposing_provider_owned_client_marks_wrapper_disposed` test exercises the provider-owns-client path (lazy client creation, `_ownsClient` via reflection, double-dispose safety, `_disposed` assertion).
- ✅ Dead `StringContainsCharMethod`/`StringStartsWithCharMethod` fields removed.
- ✅ `DynamoPartiQlStatementValidator` remediation text restored, now per-operation (write vs. read wording), backed by tests in `SaveChangesEdgeCasesTests` and `ReturnConsumedCapacityRequestTests`.
- ✅ `release-drafter.yaml`: both `release-drafter/release-drafter` and its `autolabeler` are now pinned to the peeled v7 commit SHA; job-level permissions were also tightened (workflow-level dropped to `contents: read`/`pull-requests: read`, `contents: write` scoped only to the `draft` job).
- ✅ `publish-preview.yaml`: `secrets: inherit` removal confirmed intentional — downstream template only needs the auto-injected `GITHUB_TOKEN`, no external secrets. All reusable workflow/action refs across pr-build/publish-preview/publish-release now consistently pinned to the same SHA (`76a2269c...`).

Reviewed the rest of the new commits (release-drafter `tag-filter` → `filter-by-range` migration, permission rebalancing across pr-build/publish-preview/publish-release, the `disable_autolabeler` input removal) — these look like consistent, deliberate contract alignment rather than regressions, and CI is green end-to-end including the actual build/test/publish-preview jobs.

No new correctness or safety issues found. One non-blocking suggestion left inline: a short comment near the publish-preview reusable-workflow call noting secrets aren't needed, so a future contributor doesn't reflexively re-add `secrets: inherit`.

Approving.

